PR #4, opened 2020-11-02
SIDE-931 - Add search UI services / sensors

dmorrison-olive:
https://crosschx.atlassian.net/browse/SIDE-931


scottkipferOlive (approved):

dmorrison-olive:
I'm trying to add examples for the global and sidebar searches now...

ldk/go/examples/search-global/loop/loop.go
```diff
@@ -0,0 +1,83 @@
...
+	c.sidekick = sidekick
+
+	return sidekick.UI().ListenGlobalSearch(c.ctx, func(text string, err error) {
+		c.logger.Info("controller loop callback called")
+		if err != nil {
+			c.logger.Error("received error from callback", err)
+			return
+		}
+
+		go func() {
```
dmorrison-olive, line 62: I guess we'll need to (ideally) take this wrapped goroutine out of the examples when the fix you mentioned is in place, @richseviora? Should I leave a TODO comment here?

dmorrison-olive:
I added some example loops for the global search and searchbar.

The global search one works for me, but I can't get the searchbar example loop to trigger. 🤔

ldk/go/examples/search-global/loop/loop.go
richseviora: After some thought, I'm going to leave the Whisper methods as-is. Users will be instructed that the function will be blocking; if they want to close the whisper early they need to cancel the context they're passing in.

So instead I'll be updating the examples so they're all wrapped in a goroutine.

ldk/go/examples/search-global/Makefile
```diff
@@ -0,0 +1,54 @@
+REPO := github.com/open-olive/loop-development-kit/ldk/go/example/clipboard
```
BillyMurrayOlive, line 1: I think this should be:
`REPO := github.com/open-olive/loop-development-kit/ldk/go/example/search-global`

ldk/go/examples/search-global/loop/loop_test.go
```diff
@@ -0,0 +1,43 @@
...
+	"context"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	ldk "github.com/open-olive/loop-development-kit/ldk/go"
+	loop "github.com/open-olive/loop-development-kit/ldk/go/examples/search-global/loop"
+	ldktest "github.com/open-olive/loop-development-kit/ldk/go/ldk-test"
+)
+
+func TestController(t *testing.T) {
```
BillyMurrayOlive, line 13: I think this test needs to do more. I would like to see it simulate the user entering text in the global search. Then confirming that it gets a whisper in response. 

Take a look at the whisper-confirm example as a guide. I think a staging approach like the one in that example would be good here.
Ex:
Stage 1 - Simulate user searching text
Stage 2 - Confirm whisper is received with correct content

ldk/go/examples/search-global/storage.json
```diff
@@ -0,0 +1,6 @@
+{
```
BillyMurrayOlive, line 1: This example doesn't use storage. This file is unused and should be removed.

ldk/go/examples/search-searchbar/Makefile
```diff
@@ -0,0 +1,54 @@
+REPO := github.com/open-olive/loop-development-kit/ldk/go/example/clipboard
```
BillyMurrayOlive, line 1: Should be:
` REPO := github.com/open-olive/loop-development-kit/ldk/go/example/search-searchbar`

ldk/go/examples/search-searchbar/loop/loop.go
```diff
@@ -0,0 +1,83 @@
...
+
+const (
+	backgroundColor = "#fff"
+	highlightColor  = "#651fff"
+	primaryColor    = "#666"
+)
+
+// Serve creates the new loop and tells the LDK to serve it
+func Serve() error {
+	l := ldk.NewLogger("loop-example")
```
BillyMurrayOlive, line 17: I think this should be more specific. 
`ldk.NewLogger("example-search-searchbar")`
or something like that.

ldk/go/examples/search-searchbar/loop/loop_test.go
```diff
@@ -0,0 +1,43 @@
...
+	"context"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	ldk "github.com/open-olive/loop-development-kit/ldk/go"
+	loop "github.com/open-olive/loop-development-kit/ldk/go/examples/search-searchbar/loop"
+	ldktest "github.com/open-olive/loop-development-kit/ldk/go/ldk-test"
+)
+
+func TestController(t *testing.T) {
```
BillyMurrayOlive, line 13: Same comment I made for the search-global example. This test needs more testing.
Also, please update the name to something more descriptive. `TestSearchResponse` or something like that.

ldk/go/examples/search-searchbar/storage.json
```diff
@@ -0,0 +1,6 @@
+{
```
BillyMurrayOlive, line 1: Unused and should be remove.

BillyMurrayOlive (changes requested):
Good work! I made some suggestions for improving the example tests that I would like to see implemented before this is merged.

dmorrison-olive:
Thanks for reviewing, @BillyMurrayOlive! I'll take a look at your feedback.

ldk/go/examples/search-global/Makefile
dmorrison-olive: Whoops - thanks!

Also, `github.com/open-olive/loop-development-kit/ldk/go/example` should be changed to *examples* everywhere, right?

ldk/go/examples/search-searchbar/loop/loop.go
dmorrison-olive: Changing that gives me pause as it's "loop-example" in all the examples right now, and we'd need to change it everywhere plus remember to change it when creating a new example (if we copy an existing one). Thoughts?

BillyMurrayOlive: That's a good point. I think it should be specific everywhere. We have missed it in all the other examples. You're right that it will make copy pasting more intensive, but I think it is worth the cost to have logs accurately labeled.

Consider the situation where you might have an example build from something you worked on in the past. You go to deploy a new example for testing and see a log. You assume the log is from your new example because you forgot you had the other example already build. Could mislead a person into following the wrong path.

ldk/go/examples/search-global/Makefile
BillyMurrayOlive: Oh yes. Good catch. It should be plural.

dmorrison-olive: Cool - I'll change it everywhere. 👍

ldk/go/examples/search-searchbar/loop/loop.go
dmorrison-olive: Fair enough counterpoint. That could be confusing in the logs. I'll change this and try to change it everywhere else to be specific as well. 👍

ldk/go/examples/search-global/loop/loop.go
dmorrison-olive: Hey @richseviora and @BillyMurrayOlive, I just noticed that when the call to `Whisper().Markdown()` is wrapped in a goroutine for the clipboard example, that the test always passes (I tried putting in some text to make the assertion fail, and it still worked). I assume this is because I need to put in some coordination (or however you say) between the goroutine and the test assertion? Got any ideas about how to do this?

I'm asking because this would probably apply to the tests for the global and searchbar search example loops.

dmorrison-olive: Nevermind - I finally grokked just using a channel and blocking on it until it's sent something in the fake Markdown() goroutine. 👍

dmorrison-olive:
> I added some example loops for the global search and searchbar in Go.
> 
> The global search one works for me, but I can't get the searchbar example loop to trigger. 🤔

For me now, the global and searchbar searches sporadically work (maybe 1/3 searches) when I use the samples here. I'm not sure why. 🤔

JonBernardOlive (approved):

BillyMurrayOlive:
> > I added some example loops for the global search and searchbar in Go.
> > The global search one works for me, but I can't get the searchbar example loop to trigger. 🤔
> 
> For me now, the global and searchbar searches sporadically work (maybe 1/3 searches) when I use the samples here. I'm not sure why. 🤔

BillyMurrayOlive:
> > I added some example loops for the global search and searchbar in Go.
> > The global search one works for me, but I can't get the searchbar example loop to trigger. 🤔
> 
> For me now, the global and searchbar searches sporadically work (maybe 1/3 searches) when I use the samples here. I'm not sure why. 🤔

@dmorrison-olive Can you upload what you have so far for the loop example tests.

dmorrison-olive:
> > > I added some example loops for the global search and searchbar in Go.
> > > The global search one works for me, but I can't get the searchbar example loop to trigger. 🤔
> > 
> > 
> > For me now, the global and searchbar searches sporadically work (maybe 1/3 searches) when I use the samples here. I'm not sure why. 🤔
> 
> @dmorrison-olive Can you upload what you have so far for the loop example tests?

@BillyMurrayOlive, I have everything pushed.

ldk/go/examples/search-searchbar/loop/loop_test.go
```diff
@@ -0,0 +1,48 @@
...
+	loop "github.com/open-olive/loop-development-kit/ldk/go/examples/search-searchbar/loop"
+	ldktest "github.com/open-olive/loop-development-kit/ldk/go/ldk-test"
+)
+
+func TestSearchResponse(t *testing.T) {
+	markdownDoneChan := make(chan bool)
+
+	sidekick := &ldktest.Sidekick{
+		UIService: &ldktest.UIService{
+			ListenSearchbarf: func(ctx context.Context, cb ldk.ListenSearchHandler) error {
```
DrewRepaskyOlive, line 18: this reads as "listen-search-barf" 😂

dmorrison-olive: I know, right?! haha

DrewRepaskyOlive (approved):

BillyMurrayOlive (approved):

PR #7, opened 2020-11-05
SIDE-823 - Network Sensor

scottkipferOlive:
implemented Network Service for the Go and Node LDKs.
Added HTTPRequest Interface inside the Network service to allow users to make network requests.


ldk/node/src/hostClients/networkService.ts
```diff
@@ -0,0 +1,14 @@
+export interface HttpRequest {
+  url: string;
+  method: string;
+  body: string | Uint8Array;
```
richseviora, line 4: This is a good first step; I think I'll probably end up adding support for providing objects to be JSONified, but this needn't be implemented here

proto/network.proto
```diff
@@ -0,0 +1,21 @@
...
+
+service Network {
+    rpc HTTPRequest (HTTPRequestMsg) returns (HTTPResponseMsg);
+}
+
+message HTTPRequestMsg {
+    Session session = 1;
+    string url = 2;
+    string method = 3;
+    bytes body = 4;
```
richseviora, line 15: As discussed this should include a way for consumers to supply (and read) headers, but I understand this will be implemented in a subsequent PR.

richseviora (approved):

proto/network.proto
BillyMurrayOlive: https://crosschx.atlassian.net/browse/SIDE-969

JonBernardOlive (approved):

BillyMurrayOlive (approved):

PR #8, opened 2020-11-08
[SIDE-933] Blocking Whisper Update and Bug Fixes

richseviora:
This PR:

- Updates the Node LDK to support the cancellation of whisper messages.
- Updated the Go LDK to make it clear that the Go LDK `Whisper.Markdown` and `Whisper.Content` messages do not return under the whisper is closed.
- Changes how services and connections are managed by Olive Helps and the Go LDK.
- Corrects the `package` statement in the proto file definitions for the Broker and StdIO services, which are only consumed by the Node LDK.(`github.com/hashicorp/go-plugin` supplies these for the Go LDK).

### Explanation

I thought this ticket was going to be pretty quick. I was wrong. 

After making the Node LDK changes, I first ran into issues connecting the Go Loop Client to the Node Loop Server. It turns out that the `package` instruction in the proto file is included in the service definition, and the `go-plugin` library expects the Broker and StdIO services its connects to be part of the `plugin` package.

Afterwards, I started running into a lot of non-deterministic connection issues with the Node LDK local plugins where the Node LDK was usually complaining that the `proto.Clipboard` or `proto.Whisper` services were not being served by the Olive Helps GRPC server. 

After some research and experimentation, it turns out that the way we had been initiating the services in the `LoopClient.LoopStart` had been exposing us to a race condition that only really became apparent when different GRPC implementations started connecting to enough services. We were instantiating a new GRPC server for each of the services, instead of one GRPC server for all services, and something about the goplugin client connection process hid that issue. The Node GRPC server ran into that issue head first and could not reliably connect to the servers in the right order.

So, in `LoopClient.LoopStart` we now register all the different services to one GRPC server and serve that. Similarly, in `LoopServer.LoopStart`, we only need to maintain one connection instead of several. I'm not entirely sure why this is not a similar issue in the Node LDK, right now I'm chalking that up to differences in implementation.


ldk/go/clipboardServer.go
```diff
@@ -40,7 +40,7 @@ func (m *ClipboardServer) ClipboardReadStream(emp *proto.ClipboardReadStreamRequ
 			fmt.Println("error: ldk.ClipboardServer.ClipbardReadStream -> stream.Send:", e)
 		}
 	}
-
+	//fmt.Println("Starting Clipboard Read Stream")
```
dmorrison-olive, line 43: Maybe can remove this.

ldk/node/src/hostClientFacade.ts
```diff
@@ -36,21 +39,25 @@ export default class HostClientFacade implements HostServices {
...
-      this.clipboard.connect(connInfo, session),
-      this.cursor.connect(connInfo, session),
-      this.hover.connect(connInfo, session),
-      this.fileSystem.connect(connInfo, session),
-      this.process.connect(connInfo, session),
-      this.window.connect(connInfo, session),
-      this.browser.connect(connInfo, session),
+      // this.browser.connect(connInfo, session, this.logger),
+      // this.hover.connect(connInfo, session, this.logger),
+      // this.window.connect(connInfo, session, this.logger),
```
dmorrison-olive, line 53: Safe to remove these?

ldk/go/clipboardServer.go
richseviora: Removed!

ldk/node/src/hostClientFacade.ts
richseviora: I'm actually going to comment them, and their corresponding properties out. We haven't implemented those services yet but we should being doing that sometime soon.

BillyMurrayOlive (approved):

ldk/go/loopClient.go
```diff
@@ -29,113 +30,69 @@ func (m *LoopClient) LoopStart(host Sidekick) error {
...
+		ServiceHosts: serviceHosts,
 		// TODO: Define Session here
 		Session: &proto.Session{
 			LoopID: "LOOP_ID",
 			Token:  "TOKEN",
 		},
 	})
-
+	if err != nil {
+		fmt.Println("loopClient.go ERROR FOLLOWS")
```
DrewRepaskyOlive, line 93: Seems like these 2 lines should be combined into 1.  Also, should this be using a logger?  I feel like stdout just gets eaten by Sidekick and not recorded anywhere.

ldk/node/examples/clipboard/plugin.json
```diff
@@ -0,0 +1,13 @@
+{
+  "author": "Olive",
+  "created": "2020-06-22T00:00:00Z",
+  "dependencies": [],
+  "description": "This is an example controller written in Go.",
```
DrewRepaskyOlive, line 5: Node.js example says "written in Go" and uses "controller" instead of "loop."  Also line 8 says "Example Go"

DrewRepaskyOlive (approved):

ldk/go/loopClient.go
richseviora: We should, but haven't yet, add loggers on the Olive Helps server sides. I'll make a card for that.

I don't think this should be one line, I'd like to have the first line separated from the second. These have been writing to stdout so it worked for me in the debugging process.

richseviora: SIDE-985 created

ldk/node/examples/clipboard/plugin.json
richseviora: Ah! Thanks, fixed!

ldk/go/loopClient.go
iangallOlive: @DrewRepaskyOlive I **think** OH still reads stdout and logs it, but it won't be nearly as nice as something that's using the LDK's `Logger`s, which format the messages in a better way for OH to parse. So I'd agree that using the LDK `Logger`s are the best option, but normal stdout _should_ still work, worst case scenario

iangallOlive (approved):

ldk/go/loopClient.go
DrewRepaskyOlive: If we're running from the command-line, then devs can see stdout in the console.  I just meant that when I make a call to `fmt.Printf` inside of a Loop, those messages do not show up in Sidekick log _files_, so we have to use a logger instead.  I'm more concerned with the published desktop app run by end users and not losing error context there.
Thanks for opening that ticket @richseviora

PR #9, opened 2020-11-09
Side 916 share services across loops

BillyMurrayOlive:
Previously, the session validation was being done inside the LDK on the service side of the services. But, for some services (Whisper and Storage), session information is needed for the actual implementation. 

- The Storage methods prefix the keys with the loop ID
- The whisper needs loop ID information to populate whisper footer

So, the session validation was moved from the LDK into the service implementation in Sidekick. It is passed from the LDK into Sidekick by creating a new ctx with a session value. https://golang.org/pkg/context/#WithValue

Because ctx needs to now be passed for all methods, any methods that were previously missing ctx (non-stream methods) have had ctx added.


ldk/go/Makefile
```diff
@@ -22,7 +22,20 @@ proto:
 	GO111MODULE=off go get -u github.com/golang/protobuf/protoc-gen-go
 	protoc -I ../../proto/ ../../proto/*.proto -I $(GOPATH)/src --go_out=plugins=grpc:proto/
 
-test:
+test-examples:
```
dmorrison-olive, line 25: Awesome! I was wondering if we should fire the example tests here. 👍

DrewRepaskyOlive (approved):

ldk/go/Makefile
BillyMurrayOlive: Yeah, I didn't realize the tests examples tests were not already running. Some of them were quite out of date.

scottkipferOlive (approved):

PR #15, opened 2020-11-12
SIDE-1004 - Fix ESLint warnings

dmorrison-olive:
https://crosschx.atlassian.net/browse/SIDE-1004

Warnings from ESLint are creating noise in LDK PRs. See [this PR](https://github.com/open-olive/loop-development-kit/pull/13/files) for an example.

PR #18, opened 2020-11-13
SIDE-1004 - Add template docs back

dmorrison-olive:
Addressing @richseviora's feedback [here](https://github.com/open-olive/loop-development-kit/pull/15#discussion_r523056832).

The `@template` tag [seems to be supported](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#template) for TypeScript/JSDoc.

### DIFF
--- a/ldk/node/src/hostClients/stoppables.ts
+++ b/ldk/node/src/hostClients/stoppables.ts
@@ -1,11 +1,15 @@
 /**
  * The StreamListener type is a function provided to start a stream. It accepts two parameters and returns nothing.
  *
+ * @template T - The type of the messages sent.
  * @param error - Null, unless an error is present.
  * @param input - The event that's being generated.
  */
 export type StreamListener<T> = (error: string | null, input?: T) => void;
 
+/**
+ * @template T - The type of the messages sent.
+ */
 export interface StoppableStream<T> {
   stop(): void;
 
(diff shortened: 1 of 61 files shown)


ldk/node/src/hostClients/stoppables.ts
```diff
@@ -1,11 +1,15 @@
...
  * The StreamListener type is a function provided to start a stream. It accepts two parameters and returns nothing.
  *
+ * @template T - The type of the messages sent.
  * @param error - Null, unless an error is present.
  * @param input - The event that's being generated.
  */
 export type StreamListener<T> = (error: string | null, input?: T) => void;
 
+/**
+ * @template T - The type of the messages sent.
```
richseviora, line 11: Sorry, these should be `@typeParam` tags not `@template`.

richseviora (changes requested):
Almost there :)

ldk/node/src/hostClients/stoppables.ts
dmorrison-olive: It seems to be an alias? See [here](https://typedoc.org/guides/doccomments/#typeparam-param-name) and the comment "You may also use the template tag".

`@typeParam` gives me an ESLint warning, whereas `@template` doesn't. I don't see where I can configure this either. 😕

richseviora: OH! TIL!

That `@template` is an alias for `@typeParam` is totally not intuitive at all but I'll take it!

I think you can configure whichever rule is complaining about to support additional tags in the eslint config file

richseviora (approved):

BillyMurrayOlive (approved):

PR #19, opened 2020-11-16
SIDE-979 Removed whisper style

BillyMurrayOlive:
Proto Changes
- Removed whisper style from proto definitions

Regenerate Proto
- Regenerated proto for Go LDK
- Regenerated proto for Node LDK
- [x] Need to regenerate proto for C# LDK (How do you do this?)

LDK Changes
- Removed whisper style from Go LDK
- Removed whisper style from Node LDK
- Removed whisper style from C# LDK

Example Changes
- Updated Go LDK examples
- No updates required for Node LDK examples
- No examples for C# LDK


BillyMurrayOlive:
@richseviora How do I generate the proto files for C#?

iangallOlive:
Billy edited the C# LDK; that makes him an expert!

richseviora:
> Billy edited the C# LDK; that makes him an expert!

This is the way.

@BillyMurrayOlive the C# LDK builds the GRPC files at compile time.

iangallOlive:
So this change is something I've been waiting for; I believe this would be considered a breaking change with the proto contracts (because we're _removing_ a property), right? If so, I'm concerned about how existing loops and Olive Helps apps will work. I know we haven't officially released yet and we have an auto-update feature to hopefully keep people on the latest version, but I still want to bring this up.
My thinking was this:
Once the Olive Helps repo updates to use this version of the LDK, I would expect that any loop that hasn't updated to this LDK wouldn't work. And vice versa: any loop that updates to use this version of the LDK won't work with any (old) version of Olive Helps that hasn't updated to this version of the LDK. And this gets complicated when things are in the loop library
If I'm wrong about any of that, then ignore me! Otherwise I think we need to talk about it.

iangallOlive (approved):

richseviora:
I think that's a good point; these changes are going to be in some way breaking and we should be keep tracking of all the Loops we need to update and republish whenever we make a breaking change.

In this case, I don't think that dropping the Whisper Style message will be immediately breaking but if we re-use that field it will be... in either event, as part of our launch checklist we should be checking to ensure all the Loop content remains valid and republishing as necessary. https://docs.microsoft.com/en-us/aspnet/core/grpc/versioning?view=aspnetcore-5.0

ldk/csharp/OliveHelpsLDK/OliveHelpsLDK/Whispers/WhisperClient.cs
```diff
@@ -55,21 +55,10 @@ public async Task FormAsync(WhisperForm message, CancellationToken cancellationT
 
         private static WhisperMeta GenerateMeta(WhisperConfig config)
         {
-            var styleBackgroundColor =
```
richseviora, line 58: We can also now drop the default values (L11-13)  and the coercion function (L76-79)

richseviora (changes requested):
Some C# changes but otherwise we're golden :D

ldk/csharp/OliveHelpsLDK/OliveHelpsLDK/Whispers/WhisperClient.cs
BillyMurrayOlive: Thanks for pointing that out. They have been removed.

richseviora (approved):

PR #22, opened 2020-12-01
fix for bad access: nil dereference

DrewRepaskyOlive:
Bugfix for when Loops are attempting to be gracefully shutdown, the global/searchbar on teardown were getting a fatal error: `bad access: nil dereference`

Found while debugging [SIDE-1021](https://crosschx.atlassian.net/browse/SIDE-1021)

### DIFF
--- a/ldk/go/uiClient.go
+++ b/ldk/go/uiClient.go
@@ -29,11 +29,7 @@ func (u *UIClient) ListenSearchbar(ctx context.Context, handler ListenSearchHand
 				break
 			}
 			if err != nil {
-				message := "Error with SearchbarStream"
-				if resp != nil {
-					message = resp.Text
-				}
-				handler(message, err)
+				handler("", err)
 				return
 			}
 			if resp.GetError() != "" {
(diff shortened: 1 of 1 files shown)


dmorrison-olive (approved):

iangallOlive (approved):

ldk/go/uiClient.go
```diff
@@ -29,7 +29,11 @@ func (u *UIClient) ListenSearchbar(ctx context.Context, handler ListenSearchHand
 				break
 			}
 			if err != nil {
-				handler(resp.Text, err)
+				message := "Error with SearchbarStream"
```
BillyMurrayOlive, line 32: This is admittedly a nitpick. But I feel like error information shouldn't be returned in the message.
Thoughts on the following approach?
`
if err != nil {
  if resp == nil {
    handler("", err)
  } else {
    handler(resp.Text, err)
  }
  return
}
`

DrewRepaskyOlive: I do think I prefer your way @BillyMurrayOlive.  Seems safer to have an empty message to encourage the callback to check the error.  Like the relevant details should all be in the `err` itself and not fragmented.  What if I just shorten it to the following?
```
if err != nil {
  handler("", err)
  return
}
```

BillyMurrayOlive: I think your abbreviated version is good. I was trying to accommodate returning resp.Text when you have it. But on second thought, when a function returns an error, you typically disregard the other arguments. So I don't think it is necessary.

BillyMurrayOlive (approved):

ldk/go/uiClient.go
DrewRepaskyOlive: change applied, thank you Billy!

PR #25, opened 2020-12-09
[SIDE-1076] - fix session attachment

bbrewer-oliveai:
- fixes session attachment for a few different sensors (hover, fileSystem, keyboard, etc.)
- fixes API for a few wrappers where a stream listener was not being taken
- tests are aimed at being easy to add for future services, though an integration testing strategy would be better overall for these


ldk/node/src/hostClients/keyboardClient.test.ts
```diff
@@ -0,0 +1,147 @@
...
+      sentResponse = new Messages.KeyboardTextStreamResponse().setText('hello');
+      const stream = createEmptyStream();
+
+      streamCallback = jest.fn().mockImplementation(identityCallback);
+      streamTextMock.mockImplementation(createStreamingHandler(stream));
+
+      subject.streamText(streamCallback);
+
+      sentRequest = captureMockArgument(streamTextMock);
+      stream.emit('data', sentResponse);
```
bbrewer-oliveai, line 104: This is the only place I'm testing the actual wiring of the stream  back into the provided callback, largely b/c this one was not wired up correctly (pending me reviewing the other LDKs).

ldk/node/src/hostClients/keyboardClient.ts
```diff
@@ -102,22 +102,27 @@ export default class KeyboardClient
...
+    );
     return new TransformingStream(
       this.client.keyboardHotkeyStream(message),
       transformHotKeyEvent,
       listener,
     );
   }
 
-  streamText(): StoppableStream<string> {
+  streamText(listener: StreamListener<string>): StoppableStream<string> {
```
bbrewer-oliveai, line 115: This was not taking a listener. I'll triple check this with the other LDKs, but I'm 90% certain this should take a callback.

ldk/node/src/hostClients/keyboardClient.ts
```diff
@@ -102,22 +102,27 @@ export default class KeyboardClient
...
     return new TransformingStream(
       this.client.keyboardTextStream(
         new messages.KeyboardTextStreamRequest().setSession(
           this.createSessionMessage(),
         ),
       ),
-      (response) => response.getText(),
+      (response) => {
```
bbrewer-oliveai, line 122: Will revert this.

ldk/node/src/jest.helpers.ts
```diff
@@ -0,0 +1,89 @@
+import {
```
bbrewer-oliveai, line 1: Not sure this is the best place for this file. I'm not a huge fan of DRYing up tests if it means they are less readable. However, these seemed like reasonable utility functions to me.

ldk/node/src/hostClients/keyboardClient.ts
richseviora: You are 100% correct! The stoppable streams absolutely should accept a listener.

ldk/node/src/jest.helpers.ts
richseviora: I agree with you both on the principle and the judgment call here!

ldk/node/package.json
```diff
@@ -72,6 +72,7 @@
     "grpc_tools_node_protoc_ts": "4.1.4",
     "husky": "4.3.0",
     "jest": "26.4.2",
+    "jest-create-mock-instance": "1.1.0",
```
bbrewer-oliveai, line 75: I'm not a fan of adding this, but I was struggling pretty hard to get jest and ts-jest to do this for me without complaint.

ldk/node/src/hostClients/browserClient.test.ts
```diff
@@ -0,0 +1,156 @@
...
+
+const logger = new Logger('test-logger');
+
+describe('BrowserClient', () => {
+  let subject: BrowserClient;
+  let mockGRPCClient: jest.Mocked<Services.BrowserClient>;
+  let connInfo: ConnInfo.AsObject;
+  let session: Session.AsObject;
+
+  beforeEach(async () => {
```
bbrewer-oliveai, line 32: These could be pared down further into something that just wires up a useable `subject`. I've spent too much time on this though, so I'll just stop here, which is a decent improvement IMO.

ldk/node/src/hostClients/windowClient.test.ts
```diff
@@ -0,0 +1,171 @@
...
+        createCallbackHandler(sentResponse),
+      );
+
+      queryResult = subject.queryActiveWindow();
+
+      sentRequest = captureMockArgument(mockGRPCClient.windowActiveWindow);
+    });
+
+    it('should return a transformed response', async () => {
+      await expect(queryResult).resolves.toMatchObject({
```
bbrewer-oliveai, line 66: I'd like to be able to build a helper for `toMatchInterface` as this will likely get brittle quickly, but I'm not sure the JSR (juice to squeeze ratio) is there at the moment.

ldk/node/src/test.helpers.ts
```diff
@@ -0,0 +1,127 @@
...
+) {
+  return (
+    request: TRequest,
+    metadata: Metadata,
+    options: Partial<CallOptions>,
+    callback: CallbackFunc,
+  ): ClientUnaryCall => {
+    const clientCall = new ClientUnaryCallImpl();
+
+    const actualCallback = callback || options || metadata;
```
bbrewer-oliveai, line 34: TypeScript was really busting my chops on this one. If there is a way I can define only the request + callback version and not have it gripe at me, I'm all ears

ldk/node/src/hostClients/browserClient.test.ts
```diff
@@ -0,0 +1,156 @@
...
+    ).resolves.toBeUndefined();
+  });
+
+  describe('#queryActiveURL', () => {
+    let sentRequest: Messages.BrowserActiveURLRequest;
+    let sentResponse: Messages.BrowserActiveURLResponse;
+    let queryResult: Promise<string>;
+
+    beforeEach(async () => {
+      sentResponse = new Messages.BrowserActiveURLResponse();
```
richseviora, line 53: I wondered to myself if there's value in setting an explicit value for `Url` property in the response, and then testing that that's returned but we are testing equality on line 65. I think I'm hesitating over this because GRPC primitive properties are auto-set at first.

Yeah this is probably fine :)

ldk/node/src/hostClients/browserClient.test.ts
```diff
@@ -0,0 +1,156 @@
...
+
+      subject.streamSelectedText(identityCallback);
+
+      sentRequest = captureMockArgument(
+        mockGRPCClient.browserSelectedTextStream,
+      );
+    });
+
+    it('should have attached the initial connection session to the request', () => {
+      expect(sentRequest.getSession()?.toObject()).toStrictEqual(session);
```
richseviora, line 153: I wondered if the safe navigation operator was introducing a false negative risk here, but since we're assigning `session` in the before each block, this shouldn't be an issue (and probably stops TS from complaining, though the not-null assertion operator would've worked too I think)

ldk/node/src/hostClients/hoverClient.test.ts
```diff
@@ -0,0 +1,119 @@
...
+
+    it('should call grpc client function', async () => {
+      expect(mockGRPCClient.hoverRead).toHaveBeenCalledWith(
+        expect.any(Messages.HoverReadRequest),
+        expect.any(Function),
+      );
+    });
+
+    it('should have configured the request with the right x coordinates', () => {
+      expect(sentRequest.getXfromcenter()).toBe(xFromCenter);
```
richseviora, line 80: We don't have a one assertion per test rule here. So to the extent you'd like, feel free to consolidate multiple assertions into one test where you feel it makes sense. E.g. where you can evaluate multiple like assertions into one assertion block so you're testing multiple conditions in one assertion and they're failing together.

ldk/node/src/hostClients/windowClient.test.ts
richseviora: IMO sometimes brittleness isn't the worst thing in tests; it helps us identify when something changes.

ldk/node/src/test.helpers.ts
richseviora: Ah, because each of the service calls have three signatures? Sometimes the reasonable thing to do is slap `as any` on it, especially where we're working with test code and it's not worth the effort to wrangle Typescript into behaving.

richseviora (approved):
This is really great!

ldk/node/src/test.helpers.ts
bbrewer-oliveai: I'll take another peak. Thanks!

ldk/node/src/hostClients/browserClient.test.ts
bbrewer-oliveai: Oh yeah, I was setting that for some things, but at some point, for expediency, I decided that equality would do for the moment.

ldk/node/src/hostClients/hoverClient.test.ts
bbrewer-oliveai: I'll do that. I think this one is pretty much the only place, as it was the first set I wrote. Others, like hotkey are paired together so it's NBD overall. I'll take another sweep.

ldk/node/src/hostClients/browserClient.test.ts
bbrewer-oliveai: Yeah, I can triple check that if `?.` does return early that it still throws.

ldk/node/src/test.helpers.ts
bbrewer-oliveai: `as any` did the trick. Thanks!

ldk/node/src/hostClients/browserClient.test.ts
bbrewer-oliveai: As in, in the past, sometimes I've even set values to be a `jest.fn()` to get away from magic stringiness in tests (a little extreme, yeah...).

ldk/node/src/hostClients/browserClient.test.ts
bbrewer-oliveai: Actually, for these purposes, I'll just do a not-null check. If there had been a `matchesInterface` that might have been a good middle ground.

ldk/node/src/hostClients/hoverClient.test.ts
bbrewer-oliveai: Addressed

ldk/node/src/hostClients/browserClient.test.ts
bbrewer-oliveai: Addressed

ldk/node/src/hostClients/browserClient.test.ts
bbrewer-oliveai: I've thought about this some more, and I'll actually set explicit values. I agree that there are likely places where this might pass incorrectly.

bbrewer-oliveai: Addressed

BillyMurrayOlive (approved):
I don't know much about jest, but it looks reasonable. I downloaded the branch ran the tests. After realizing that I needed to update dependencies, everything ran great.

PR #27, opened 2020-12-10
[SIDE-1045] Correct linting rule, remove unnecessary disables

richseviora:
This PR fixes a bug with our linting configuration that was generating false positives on the `no-shadow` rule. It turns out that that rule does not work with Typescript files, and we need to use the one from `@typescript-eslint` instead.

Also fixes minor linting complaints in `src/brokerGrpcServer.ts` file.


bbrewer-oliveai (approved):
Just makes sense

iangallOlive (approved):

PR #28, opened 2020-12-10
Side 969 network sensor add support for headers

JonBernardOlive:
Also added a network sensor example to be able to test the network sensor with.

Screen Shot 2020-12-10 at 4 08 41 PM


ldk/go/examples/network/readme.md
```diff
@@ -0,0 +1,59 @@
+# Sidekick Loop Example in Go
+
+This is a very simple loop to use as a starting point. This loop watches for text events and emits a whisper every time it receives an event.
```
bbrewer-oliveai, line 3: Is this true or does it send a network request?

bbrewer-oliveai (commented):
My go is a bit rusty, but this makes sense to me. I'll comment for now, but can approve if you want.

ldk/go/examples/network/Makefile
```diff
@@ -0,0 +1,54 @@
...
+	GOOS=windows go build  \
+	     -ldflags "-X ${REPO}/loop.GitBranch=${GIT_BRANCH} -X ${REPO}/loop.GitCommit=${GIT_COMMIT} -X ${REPO}/loop.Version=${VERSION}" \
+	     -o build/windows-amd64/plugin.exe ./
+
+clean:
+	rm -rf ./build
+
+build: clean build/darwin-amd64/plugin build/windows-amd64/plugin
+
+# TODO: remove before merging to master
```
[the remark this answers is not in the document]
JonBernardOlive: This was copied from another example... So 🤷‍♂️

ldk/go/examples/network/readme.md
JonBernardOlive: This was copied from another example... So 🤷‍♂️ I'll look into updating this.

proto/network.proto
```diff
@@ -13,9 +13,11 @@ message HTTPRequestMsg {
     string url = 2;
     string method = 3;
     bytes body = 4;
+    map <string, string> headers = 5;
 }
 
 message HTTPResponseMsg {
     uint32 responseCode = 1;
     bytes data = 2;
+    bytes headers = 3;
```
richseviora, line 22: Can we discuss this? I’m really uncomfortable passing the Response Header as a JSON object serialized to `bytes` instead of a `map <string,string>` like request headers. At some point they will need to be deserialized to key-value pairs before they're exposed to LDK consumers. At the very least, I'd like to understand the considerations that drove this decision.

JonBernardOlive: I did it this way because I couldn't find a way to convert the resulting headers from the `http.Header` type to `map[string][]string` so it can be passed through the proto, probably because I'm dumb.

JonBernardOlive: Let me tinker with it a little bit more to see if I can get it looking better

JonBernardOlive: I was hoping to not have to do something like this to achieve this.
`	responseHeaders := make(map[string][]string)

	for name, value := range resp.Header {
		responseHeaders[name] = value
	}`

JonBernardOlive: I'm open to suggestions

richseviora: TBH I think there'll be a bit of tedium here to convert the headers from the Go map of string arrays that's provided for in the `net/http/Header` type to a proto `map<string, string>`. This is the sort of thing that would be easy to represent if Go supported generics, but since it doesn't we gotta roll these mapping functions ourselves. So it's what you said above, but also converting the value from `[]string` to `string` (seperated by commas)

bbrewer-oliveai (approved):
Overall, I'm able to get the new headers field back from OliveHelps with this and the headers definition for request and response make sense to me.

ldk/go/examples/network/Makefile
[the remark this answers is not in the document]
BillyMurrayOlive: Yeah, these are in pretty much every example. This command is real useful.

ldk/go/networkClient.go
```diff
@@ -12,17 +13,19 @@ type NetworkClient struct {
 	session *Session
 }
 
-// HTTPResponse is the structure recieved from HttpRequest
+// HTTPResponse is the structure received from HttpRequest
 type HTTPResponse struct {
 	ResponseCode int
 	Data         []byte
+	Headers      map[string]*structpb.ListValue
```
BillyMurrayOlive, line 20: We shouldn't return a protobuf type to the developer. Can we call `*structpb.ListValue.String()` to get a string and then return a map[string]string? 

Why are we using a `*structpb.ListValue` at all? Couldn't we just send a map of strings back the same as we send?

BillyMurrayOlive: Ok, I see the other conversation now. I agree that `map[string][]string` is a mess. But I think we should return the same format that the standard http method returns.

ldk/go/examples/network/loop/loop.go
```diff
@@ -0,0 +1,188 @@
...
+	Zip            string `json:"postal_code"`
+}
+
+type apiResponse struct {
+	Meta    interface{} `json:"meta"`
+	Results []recall    `json:"results"`
+}
+
+// Serve allows Olive Helps to have access the loop
+func Serve() error {
```
BillyMurrayOlive, line 37: Thanks for adding an example!

ldk/go/networkClient.go
JonBernardOlive: "standard http method returns" implies going back to using `http.Header` and marshalling the data. Is that what you're saying here?

JonBernardOlive: @BillyMurrayOlive ^

JonBernardOlive: I tried doing `map[string][]string` but it really needs to be more like `map[string][]*proto.someRandomListDefinitionInProto`, and I've tried a couple different ways and I can seem to access these proto generated structs in sidekick.

JonBernardOlive: Does someone know a little more about that and can point me to an example?

BillyMurrayOlive (approved):
Looks great! Thanks for making those changes.

PR #29, opened 2020-12-14
[SIDE-969] - network sensor headers in C# and NodeJS clients

bbrewer-oliveai:
C# and NodeJS were verified against a stub with an integration test. Feel free to ask me about that part.


ldk/csharp/OliveHelpsLDK/OliveHelpsLDK/Network/NetworkClient.cs
```diff
@@ -29,18 +32,43 @@ public Task<HTTPResponse> HTTPRequest(HTTPRequest request, CancellationToken can
...
             };
+
+            clientResponse.Headers = ParseHeaders(response.Headers);
+
+            return clientResponse;
+        }
+
+        private static IDictionary<string, IList<string>> ParseHeaders(MapField<string, ListValue> headers)
+        {
+            var parsedHeaders = new Dictionary<string, IList<string>>();
```
bbrewer-oliveai, line 58: I can do these as the `Linq` based variants that do `map`, `filter`, `reduce`, if desired. Those are debatably more idiomatic, though I doubt it matters for internal code.

ldk/node/src/hostClients/networkClient.ts
```diff
@@ -1,8 +1,55 @@
...
+ * @param headersMap - A map of headers
+ * @internal
+ */
+function parseHeadersMap(
+  headersMap: jspb.Map<string, structpb.ListValue>,
+): Map<string, Array<string>> {
+  const headerEntries: Array<[
+    string,
+    Array<string>,
+  ]> = headersMap.getEntryList().map(([key, value]) => {
```
bbrewer-oliveai, line 28: There is actually a bug with this. This and some of the other functions actually turn the ListValues straight into Array<string> but don't reflect that in their typing. Confirmed this breaks with both a GripMock setup and the real thing coming from OliveHelps

bbrewer-oliveai: Addressed

ldk/csharp/OliveHelpsLDK/OliveHelpsLDK/Network/NetworkClient.cs
richseviora: I agree - this is fine as-is

ldk/node/src/hostClients/networkService.ts
```diff
@@ -2,11 +2,14 @@ export interface HttpRequest {
...
   method: string;
   body: string | Uint8Array;
+
+  headers: Record<string, string>
 }
 
 export interface HttpResponse {
   statusCode: number;
   data: string | Uint8Array;
+  headers: Record<string, Array<string>>;
```
richseviora, line 12: I'm wondering if it'd make more sense to return the same header value type as we sent (a single string with comma separated values instead of an array of values).

richseviora (approved):
Looks good! One thing to consider, otherwise this is great! :)

ldk/node/src/hostClients/networkService.ts
bbrewer-oliveai: We went the other way with it, but great callout.

proto/network.proto
```diff
@@ -6,19 +6,20 @@ import "google/protobuf/struct.proto";
 import "session.proto";
 
 service Network {
-    rpc HTTPRequest (HTTPRequestMsg) returns (HTTPResponseMsg);
+  rpc HTTPRequest (HTTPRequestMsg) returns (HTTPResponseMsg);
```
bbrewer-oliveai, line 9: this will be done as part of JB's work

BillyMurrayOlive (approved):
Looks good!

PR #30, opened 2020-12-15
Example integration tests run against a happy path stub

bbrewer-oliveai:
Stub config and setup is in https://github.com/bbrewer-oliveai/loop-development-kit-stub


bbrewer-oliveai:
Closing, as this is just around for reference

PR #31, opened 2020-12-15
Side 1060 disambiguation card type frame

JonBernardOlive:
The go portion. Node/c# coming in a new branch


bbrewer-oliveai (approved):
The only thing that jumps out at me on this one is that I don't have the background to know what disambiguation means in this context (I don't know if I was there for the refinement of this part). Hopefully, Rich and I will be prepping people on that, but just worth noting.

iangallOlive (approved):

richseviora:
@bbrewer-oliveai the disambiguation card type is a collection of results that the user can click on. Clicking on a result triggers a command back to us.

I'd really prefer that we call it something like "ResultsList" or the like.

JonBernardOlive:
> @bbrewer-oliveai the disambiguation card type is a collection of results that the user can click on. Clicking on a result triggers a command back to us.
> 
> I'd really prefer that we call it something like "ResultsList" or the like.

@richseviora @bbrewer-oliveai It was named this way because that's what the product team wanted to call it. If that is an issue for you guys, we'll need to have the product team to come up with a new name.

PR #32, opened 2020-12-16
SIDE-1090 - better error handling in LDKs

bbrewer-oliveai:
Adds client side, low-level (connection failure, message structure problems, etc.) error logging to each LDK client.


ldk/node/src/hostClients/fileSystemClient.int.test.ts
```diff
@@ -0,0 +1,44 @@
+import { FileSystemClient } from './fileSystemClient';
```
bbrewer-oliveai, line 1: will remove

ldk/node/src/hostClients/whisperClient.int.test.ts
```diff
@@ -0,0 +1,32 @@
+import FileSystemClient from './whisperClient';
```
bbrewer-oliveai, line 1: will remove

richseviora (approved):
This looks really good!

Can you add a comment with what these sort of errors look like in the log? I'll probably draw back a bit on the logging in my Filesystem PR

bbrewer-oliveai:
> This looks really good!
> 
> Can you add a comment (in the PR) with what these sort of errors look like in the log? I'll probably draw back a bit on the logging in my Filesystem PR

@richseviora and anyone else
Example of what this comes out as, say for example, if the session is missing from the request - it uses the logger provided by the Loop. The low-level error is also passed on to the consuming Loop as a reject or error part of the streaming callback (NodeJS, for example), and won't crash the loop, so Loop users can still deal with it gracefully if desired. C# is one exception where users will need to catch the exception still.

The only notable things about the log are the error message itself, and then the service and method that failed.
```shell
2021-01-06T14:15:58-05:00 ERR Client exception error="rpc error: code = Unknown desc = proto session is nil pointer" @module=example-network component=loop metadata={"appId":"d9d527bb-7440-490e-80da-201d0666aca2","author":"Ben Brewer","authorId":"5fce6f8c0f91ea0076089380","colorStrip":"#73ff57","created":"2021-01-05T23:52:25.08Z","description":"Description","documentation":"","entrypoint":[],"loopLibraryLogo":"","loopWhisperLogo":"","name":"network","orgId":"Olive","organization":"Olive","semverMac":"1.0.0","semverWindows":"","specification":"1","updated":"2021-01-05T23:52:25.08Z","uploadId":"","version":""} method=HTTPRequest service=Network timestamp=2021-01-06T14:15:58.063-0500
```

ldk/csharp/OliveHelpsLDK/OliveHelpsLDK/ExceptionLoggingInterceptor.cs
```diff
@@ -0,0 +1,146 @@
...
+                return action.Result;
+            });
+        }
+
+        private static IAsyncStreamReader<TResponse> WrapStreamHandler<TResponse>(IAsyncStreamReader<TResponse> call, ILogger logger)
+        {
+            return new AsyncStreamReaderWrapper<TResponse>(call, logger);
+        }
+
+        private class MethodExceptionLogger<TRequest, TResponse> : ILogger where TRequest : class where TResponse : class
```
bbrewer-oliveai, line 75: `WithFields` seems like a less bulky alternative to this that I'll do

ldk/csharp/OliveHelpsLDK/OliveHelpsLDK/ExceptionLoggingInterceptor.cs
```diff
@@ -0,0 +1,146 @@
...
+        private static Task<TResponse> WrapContinueHandler<TResponse>(Task<TResponse> task, ILogger logger)
+        {
+            return task.ContinueWith((action) =>
+            {
+                if (action.Exception == null) return action.Result;
+                action.Exception.Handle((exception) =>
+                {
+                    if (exception is RpcException)
+                    {
+                        logger.Error(exception.Message);
```
bbrewer-oliveai, line 61: I'm going to have this actually log EVERY exception and then see if there is a way to propagate them without losing them inside other continuations, like the LoggedParser, etc.

ldk/go/exceptionLoggingInterceptor.go
```diff
@@ -0,0 +1,53 @@
...
+		eli.logError(method, err)
+	}
+
+	return err
+}
+
+func (eli *ExceptionLoggingInterceptor) NewStream(ctx context.Context, desc *grpc.StreamDesc, method string, opts ...grpc.CallOption) (grpc.ClientStream, error) {
+	stream, err := eli.interceptedConn.NewStream(ctx, desc, method, opts...)
+	if err != nil {
+		eli.logError(method, err)
```
bbrewer-oliveai, line 27: no clue how gofmt was okay with this :/

BillyMurrayOlive: What's wrong with it. I must be missing something.

BillyMurrayOlive (approved):
I only feel like I can be any use for approving the Go piece of this. I approved the PR, but you should really consider it 1/3 of an approval.

ldk/go/exceptionLoggingInterceptor.go
bbrewer-oliveai: Oh yeah, it's indenting really far, but then I realized GoLang uses tabs. My bad

BrettMugglinOlive (approved):

PR #33, opened 2020-12-16
Side 1036/file system

scottkipferOlive:
- Go Implementation of Filesystem


ldk/go/cursorClient.go
```diff
@@ -36,25 +36,23 @@ func (c *CursorClient) ListenPosition(ctx context.Context, handler ListenPositio
 		return err
 	}
 
-	go func() {
```
bbrewer-oliveai, line 39: Was this just broken before?

ldk/go/file.go
```diff
@@ -0,0 +1,13 @@
+package ldk
+
+import "os"
+
+// File this is a file
+type File interface {
+	Read([]byte) (int, error)
```
bbrewer-oliveai, line 7: I've been looking at what is standard in C# and this _should_ be close enough to the `FileStream` class and the `System.IO.Stream` abstract class, though we'll likely have to make a custom interface for this as there is no actual `FileStream` interface (just a class).

ldk/go/filesystemService.go
```diff
@@ -2,25 +2,54 @@ package ldk
...
 
 	"github.com/open-olive/loop-development-kit/ldk/go/proto"
 )
 
 // FilesystemService is an interface that defines what methods plugins can expect from the host
 type FilesystemService interface {
-	Dir(context.Context, string) ([]FileInfo, error)
+	Dir(context.Context, string) ([]os.FileInfo, error)
 	ListenDir(context.Context, string, ListenDirHandler) error
-	File(context.Context, string) (FileInfo, error)
```
bbrewer-oliveai, line 14: I noticed that the "one time query for FileInfo" function is missing now. Is that basically just moved to the `Stat` on the `File` interface?

ldk/go/filesystemService.go
```diff
@@ -2,25 +2,54 @@ package ldk
...
 	ListenFile(context.Context, string, ListenFileHandler) error
+	Open(context.Context, string) (File, error)
+	Create(context.Context, string) (File, error)
+	MakeDir(context.Context, string, uint32) error
+	Copy(context.Context, string, string) error
+	Move(context.Context, string, string) error
+	Remove(context.Context, string, bool) error
 }
 
 type FileInfo struct {
```
bbrewer-oliveai, line 24: Is this the same `FileInfo` that `Stat` returns or different? Might be a bit confusing?

bbrewer-oliveai (approved):
I can't speak too much to the actual GoLang code, but the interface for the File itself makes sense to me.

ldk/go/filesystemService.go
scottkipferOlive: Correct.  You can get the FileInfo by opening a file and then calling `.Stat()`

ldk/go/file.go
BillyMurrayOlive: That's good news!

ldk/go/cursorClient.go
BillyMurrayOlive: I think so.
The problem with these streaming calls is that, if the function returns, then gRPC closes the stream. So we need the for loop that polls for data to be blocking.

BillyMurrayOlive: One thing that we need to look into when we have time is if this behavior can be configured differently. There are lots of options for these gRPC connections and I think it would be better if it didn't automatically close the stream.
https://godoc.org/google.golang.org/grpc#DialOption

ldk/go/filesystemService.go
bbrewer-oliveai: 👍

ldk/go/file.go
bbrewer-oliveai: The only other thing that jumps out at me here is that C# has ReadAsync, WriteAsync as well, which seem to fit better with an event loop (and in NodeJS, they will pretty much have to be async calls, due to the single thread). I don't know my GoLang enough to know if that (async v synch) is basically what you get with the above implicitly or not?

BillyMurrayOlive: I think the approach here in Go is closer to ReadAsync, WriteAsync. 

If would consider reading the whole file in a singe operation to the non-async analog in Go.

ldk/go/filesystemService.go
scottkipferOlive: Its the same

proto/filesystem.proto
```diff
@@ -56,22 +71,121 @@ message FilesystemDirStreamResponse {
...
+        Create create = 1;
+        Open open = 2;
+        Read read = 3;
+        Write write = 4;
+        Chmod chmod = 5;
+        Chown chown = 6;
+        Stat stat = 7;
+        Close close = 8;
+    }
+
```
bbrewer-oliveai, line 128: Should this also have the session in it like the other streams do?

ldk/go/filesystemClient.go
```diff
@@ -156,16 +132,401 @@ func (f *FilesystemClient) ListenFile(ctx context.Context, path string, handler
...
+			default:
+				return 0, errors.New("unexpected response from server")
+			}
+
+		}
+	}
+
+}
+
+func (f *GRPCFile) Read(b []byte) (int, error) {
```
DrewRepaskyOlive, line 209: I am likely calling this the wrong way, but I am getting `panic: runtime error: slice bounds out of range [:512] with capacity 0` inside this func.  [This is our Loop code](https://bitbucket.org/crosschx/loop-memorialcare-navigator/branch/SIDE-1035-memorial-care-loop-must-use-part2?dest=master#Lloop/loop.goT82).

What is the appropriate way to call this func?  I'm trying to go from `sidekick.Filesystem().Open()` which returns `ldk.File` to the actual file contents.

BillyMurrayOlive: @DrewRepaskyOlive Don't do the thing you are doing where you read it to a byte array. The `ldk.File` type is an `io.Reader`. So you can just pass it directly into `excelize.OpenReader` (https://godoc.org/github.com/360EntSecGroup-Skylar/excelize#OpenReader).

ldk/go/filesystemServer.go
```diff
@@ -2,18 +2,24 @@ package ldk
...
 )
 
 // FilesystemServer is used by the controller plugin host to receive plugin initiated communication
 type FilesystemServer struct {
 	Impl FilesystemService
 }
 
-// list the contents of a directory
+// ErrNoFile File is not initiated
+var ErrNoFile = errors.New("no file")
```
BillyMurrayOlive, line 20: Can you please make this error more verbose. Something like:
`received invalid request to act on a file before a file has been opened or created`

BillyMurrayOlive (approved):

bbrewer-oliveai:
Updates for EOF look good to me

richseviora:
@scottkipferOlive reminder this is still open - I'm going to start the Node and .Net LDK implementation work in a branch off this branch

richseviora:
I think we have a race condition with the current structure.

Right now we send a request to the stream and then read from the stream. There's no explicit linkage between the request and responses. It generally plays out like this:

Write -> File Info
Read <- File Info
Write -> Copy
Read <- Copy

The client sends a message. And then it reads the next message from the stream and assumes it's for this request. If the client code is structured in such a way that all requests are sequential then we have no problem. However, suppose this happens:

Write -> Write
Write -> FileInfo
Write -> Chown

The `FileSystemServer` is _currently_ blocking on these requests and will handle them in order they were sent. Meanwhile, the LDK has multiple threads/tasks/listeners that are each waiting to consume the next message off the stack. I don't think we can assume that the LDKs will always handle each response handler in the order they were created. If the server ever became non-blocking then we could probably assume that a Chown operation would be faster than a Write operation. So we'd have this scenario:

Write -> Write
Write -> Chown
Read (expecting Write) <- Chown
Read (expecting Chown) <- Write

This will then cause an error in the Go LDK.

This is probably a low priority issue at the moment. I suspect that most of the requests will be consecutive and are likely to be processed FIFO in any event. However, I think I'd like to solve this problem before it happens.

If we agree that this is something we want to solve now, I'd suggest the following solution:

- LDK generates a unique request ID and sends it with the request.
- OH sends the unique request ID back with its response.
- LDK receives that request and sends it back to the listener waiting for it.

bbrewer-oliveai:
RE: @richseviora 's suggestion - so this would basically be like a unique file handle, right? That makes sense to me on a conceptual level at least: in most languages you read/write/delete, etc. on a unique file handle.

richseviora:
Conversation recap:

We (myself, Ben, Scott, Billy, JB, Ian) met to discuss this. Outcomes:

- We agree this is not an issue at this time.
- There are a variety of approaches we could take to address this (LDK consumer queuing) in each of the LDKs.
- Because of the nature of the expected file system changes, we're fine instructing LDK consumers to make requests sequentially for each file.
- In the future if we decided that these changes should not be blocking and/or create another stream where we cannot guarantee that each request is followed by its response, we will assign each request an ID.

ldk/go/filesystemClient.go
DrewRepaskyOlive: Thanks @BillyMurrayOlive, that worked!

DrewRepaskyOlive (approved):

ldk/go/filesystemServer.go
```diff
@@ -181,3 +154,249 @@ func (f *FilesystemServer) FilesystemFileStream(req *proto.FilesystemFileStreamR
...
+		case <-stream.Context().Done():
+			return nil
+		default:
+			request, err := stream.Recv()
+			if err != nil {
+				return err
+			}
+
+			switch req := request.RequestOneOf.(type) {
+			case *proto.FilesystemFileStreamRequest_Open_:
```
richseviora, line 178: We're missing a branch for the Close request. Something like:

```suggestion+2
			case *proto.FilesystemFileStreamRequest_Close_:
			   return nil
```

PR #34, opened 2020-12-20
Side 1060 disambiguation card type frame node

JonBernardOlive:


BillyMurrayOlive (approved):

ldk/node/src/hostClients/whisperMessageBuilder.ts
```diff
@@ -236,11 +238,56 @@ export const generateWhisperListElement = (
   return msg;
 };
 
+export const generateWhisperDisambiguationElement = (
+  element: WhisperDisambiguationElements,
+): messages.WhisperDisambiguationElement => {
+  const WDE = messages.WhisperDisambiguationElement;
```
richseviora, line 244: Instead of assigning as a constant here, perhaps just create it with the namespace?
```suggestion
  const msg = new messages.WhisperDisambiguationElement();
```

richseviora: Though if you're doing this to save typing out the whole class each time, maybe assign it outside of the function?

IDK, I have a preference against assigning the same variable repeatedly inside a function if it's not changing inside the function, but that's just my preference.

richseviora (approved):

ldk/node/dist/hostClients/networkClient.js
```diff
@@ -60,7 +60,9 @@ class NetworkClient extends baseClient_1.default {
             msg.setUrl(req.url);
             msg.setMethod(req.method);
             msg.setBody(req.body);
-            addHeadersToMessage(msg, req.headers);
+            if (req.headers) {
```
bbrewer-oliveai, line 63: Good catch!

ldk/node/examples/whisper-disambiguation/src/index.ts
```diff
@@ -0,0 +1,182 @@
...
+import { HostServices, Logger, Loop, serveLoop } from '../../../dist';
+import {
+  WhisperDisambiguationElements,
+  WhisperListStyle,
+  WhisperListAlign,
+} from '../../../dist/hostClients/whisperService';
+
+import { decodeRecall, Recall, RecallJSON } from './transform'
+
+const logger = new Logger('olive-helps-node-example-network');
```
bbrewer-oliveai, line 13: Looks like the old logger name got copy pasta'd

bbrewer-oliveai (approved):
Glad to see more examples being added in. I'm not gonna worry too much about the linting issues, especially for just the examples. Otherwise, there was a bit of copy paste cruft between the examples, but nothing I'd consider a show stopper.

ldk/node/src/hostClients/whisperMessageBuilder.ts
JonBernardOlive: This is how the `generateWhisperListElement()` function is set up, but I agree with you. I didn't want to change anything just in case I caused some sort of unforeseen issue with how this works.

PR #35, opened 2020-12-21
update NodeJS examples to include relevant command

bbrewer-oliveai:
The redundant parts of this can be collapsed if desired. However, for the moment, duplication doesn't seem too unwieldy.

### DIFF
--- a/ldk/node/examples/clipboard/package.json
+++ b/ldk/node/examples/clipboard/package.json
@@ -4,6 +4,7 @@
   "description": "",
   "main": "src/index.ts",
   "scripts": {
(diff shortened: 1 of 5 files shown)


dmorrison-olive (approved):

richseviora:
I wonder if it'd be easier to have them run `./node_modules/.bin/ts-node src/index.ts` instead of going through the build process

bbrewer-oliveai:
> I wonder if it'd be easier to have them run `./node_modules/.bin/ts-node src/index.ts` instead of going through the build process

I like it, I'll make it so

iangallOlive (approved):

PR #36, opened 2020-12-21
update test struct to reflect updated interface

bbrewer-oliveai:


bbrewer-oliveai:
Thought I couldn't find this in JB's PR, but it's there

PR #39, opened 2020-12-24
[SIDE-1153] Publish 2.0.0-alpha

richseviora:
Version number bump to 2.0.0-alpha


bbrewer-oliveai (approved):
It just makes sense

richseviora:
One note: Right now we have a 0.1.0 release in this repo. But since we're publishing to npm and we already have v1.3.0 there this needs to be v2

scottkipferOlive (approved):

PR #41, opened 2020-12-29
[SIDE-1091] Add support for IAsyncEnumerable

richseviora:
Adds support for consuming Streaming Calls as `IAsyncEnumerable`.

### DIFF
--- a/ldk/csharp/OliveHelpsLDK/Example/Program.cs
+++ b/ldk/csharp/OliveHelpsLDK/Example/Program.cs
@@ -65,11 +65,10 @@ namespace Example
             _clipboardStream = _services.Clipboard.Stream();
             Task.Run(async () =>
             {
-                while (await _clipboardStream.MoveNext())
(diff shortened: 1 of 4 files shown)


iangallOlive (approved):

ldk/csharp/OliveHelpsLDK/Example/Program.cs
```diff
@@ -65,11 +65,10 @@ private void ClipboardStream()
             _clipboardStream = _services.Clipboard.Stream();
             Task.Run(async () =>
             {
-                while (await _clipboardStream.MoveNext())
+                await foreach (var clipboardContent in _clipboardStream.ToAsyncEnumerable())
```
bbrewer-oliveai, line 68: I like this way of doing it - gives a bit more flexibility. I could see folks being a little confused or having to use code completion to get to this point, but I think with a good example like this it's fine.

bbrewer-oliveai (approved):

PR #45, opened 2021-01-12
Sensor doc example

bbrewer-oliveai:


bbrewer-oliveai:
Closing b/c this is an example

PR #46, opened 2021-01-12
Feature/side 1189 fix dependabot security issues

bbrewer-oliveai:
Sadly, this doesn't quite get highlight.js bumped b/c through a long chain of dependencies `grpc_tools_node_protoc_ts` still wants the deprecated version of it.


iangallOlive (approved):

iangallOlive:
Thanks for taking this over and resolving the issues I had.

A couple interesting things:
- It looks like people have already reported issues on it: https://github.com/helpers/helper-markdown/issues/14
- It looks like someone has already created a PR for it (but of course no reviews/merge): https://github.com/helpers/helper-markdown/pull/16
- When I run `npm audit` on the main branch, I don't even see this being reported. I was hoping `npm audit fix` could resolve this.

ldk/node/src/hostClients/baseClient.ts
```diff
@@ -101,7 +101,7 @@ export default abstract class BaseClient<THost extends CommonHostServer>
       callback: (err: grpc.ServiceError | null, response: TResponse) => void,
     ) => void,
     builder: () => TMessage,
-    renderer: (response: TResponse) => TOutput | undefined,
+    renderer: (response: TResponse) => TOutput | PromiseLike<TOutput>,
```
richseviora, line 104: What prompted this change?

ldk/node/src/hostClients/browserService.ts
```diff
@@ -1,11 +1,15 @@
 import { StoppableStream, StreamListener } from './stoppables';
 
-export interface BrowserSelectedTextResponse {
+export interface IBrowserSelectedTextResponse {
   text: string;
   url: string;
   tabTitle: string;
 }
 
+export type BrowserSelectedTextResponse =
```
richseviora, line 9: I see we're doing this in a few places. If we think that sometimes `undefined` values are being returned I'd like to understand when that can happen, and whether it makes sense fo surface these Response Or Undefined types at this level as opposed to within the `BaseClient` or some other level of abstraction.

BrettMugglinOlive (approved):

richseviora (approved):

ldk/node/src/hostClients/baseClient.ts
bbrewer-oliveai: Explained out of band, but for the record, adding `PromiseLike<TOutput>` was me not understanding the problem at all.

For everyone else's convenience:
- PR that broke this in `typescript` itself: https://github.com/microsoft/TypeScript/pull/39817
- How other big consumers are dealing with it (look at related commits): https://github.com/microsoft/vscode/issues/106358

bbrewer-oliveai: I've updated the PR to just not go that far ahead with the update for typescript (a patch version bump instead of minor version bump), as it really wasn't the point of this PR anyways.

ldk/node/src/hostClients/browserService.ts
bbrewer-oliveai: Yeah, how I did here is just plain wrong, in hindsight. Honestly, what we likely should be doing is rejecting if something comes back as undefined in most of these cases. As in, gRPC will always set things to "optional", but at a layer above that, we have cases where we really do consider a field required enough for it to be an error. In any case, we'll deal with this when we're better prepared - not bumping the minor version for the moment.

PR #47, opened 2021-01-13
SIDE-1194 Update documentationbased on feedback from adaptrack

BrettMugglinOlive:


ldk/node/src/hostClients/networkService.ts
```diff
@@ -8,7 +8,15 @@ export interface HttpRequest {
 
 export interface HttpResponse {
   statusCode: number;
-  data: string | Uint8Array;
+  /**
```
bbrewer-oliveai, line 11: Does this show up in the generated docs and code completion? If you don't know right off-hand I can pull it down and see.

ldk/node/src/hostClients/networkClient.ts
```diff
@@ -64,7 +64,7 @@ export class NetworkClient
       },
       (response: messages.HTTPResponseMsg) => ({
         statusCode: response.getResponsecode(),
-        data: response.getData(),
+        data: response.getData_asU8(),
```
bbrewer-oliveai, line 67: @richseviora @BillyMurrayOlive - this is consistent with how we've done this in each LDK. Is there a reason we chose to do it this way instead of a decoded string? Do we expect that most partners won't typically decode the body anyways?

bbrewer-oliveai (approved):
Some curiosity around "why a byte string/array/etc." but otherwise looks good.

ldk/node/src/hostClients/networkClient.ts
bbrewer-oliveai: Just chatted with Rich and I forgot that sometimes files are NOT just text 😀 Just makes sense

ldk/node/src/hostClients/networkService.ts
BrettMugglinOlive: I think it does

ldk/node/src/hostClients/networkClient.test.ts
```diff
@@ -58,7 +58,7 @@ describe('NetworkClient', () => {
 
     beforeEach(async () => {
       sentResponse = new Messages.HTTPResponseMsg()
-        .setData('response data')
+        .setData(new Uint8Array([173, 235, 41, 162, 123, 30, 117, 171, 90]))
```
richseviora, line 61: It's not clear how this value is derived - maybe we should start with text, decode into uint8 explicitly?

richseviora (changes requested):
One minor change - other looks great :)

ldk/node/src/hostClients/networkClient.test.ts
BrettMugglinOlive: What if I add an encoder and just encode the message? With the changes below, it seems that it would always return as a uint8 array

richseviora (approved):

PR #48, opened 2021-01-14
SIDE-1090 - ensure interceptor returns results on success instead of cancelling for C#

bbrewer-oliveai:


ldk/csharp/OliveHelpsLDK/OliveHelpsLDK/ExceptionLoggingInterceptor.cs
```diff
@@ -162,7 +162,7 @@ private static void HandleException(AggregateException aex, ILogger logger, Canc
                 aex.Handle(exception =>
                 {
                     logger.Error("Client exception", exception);
-                    return true;
+                    return false;
```
bbrewer-oliveai, line 165: This makes 0 difference to everything BUT the duplex streams - it mostly means that the TaskScheduler will re-throw the exception, which is the desired outcome.

richseviora (approved):

iangallOlive (approved):

PR #50, opened 2021-01-22
[SIDE-1186] Create Self-Test Loop

richseviora:
This PR will:

- Create a testing framework in each language for the LDK.
- Test various core capabilities.


BrettMugglinOlive (approved):

ldk/csharp/OliveHelpsLDK/SelfTestLoop/TestBuilder.cs
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        public ITest[] BuildTests(ILoopServices services)
+        {
+            var testServices = new TestServices()
+            {
+                Logger = Logger,
+                Services = services,
+            };
+            var testClass = new WhisperTests() {Services = testServices};
```
bbrewer-oliveai, line 27: This is probably the only major gripe I have with this. I tried adding an extra test file and did not realize the test harness didn't add them in for me for free. Not a show stopper, but that would be nice if possible.

bbrewer-oliveai (approved):

ldk/csharp/OliveHelpsLDK/SelfTestLoop/TestBuilder.cs
richseviora: Right! OK, so that's a bit more work; I have to figure out how I supply the TestServices object... maybe I supply TestServices if it implements an interface with a settable property for the same?

Or maybe we just say that the test methods can take TestServices as an argument.

What I'll do is proceed with the merge as-is, and add that in with the additional tests

PR #51, opened 2021-01-22
Side 1244 Add links to list whispers

BrettMugglinOlive:


dmorrison-olive (approved):

ldk/csharp/OliveHelpsLDK/OliveHelpsLDK/Whispers/List.cs
```diff
@@ -23,22 +23,30 @@ public enum Align
...
 
     public class ListMessage : ListBase
     {
-        public string Header { get; }
-        public string Body { get; }
-        public Style Style { get; }
-        public Align Align { get; }
+        public string Header { get; set; }
+        public string Body { get; set; }
+        public Style Style { get; set; }
```
bbrewer-oliveai, line 36: As we discussed, these never should have worked, as you can't really mutate a string through a getter like you could a List or Dictionary

ldk/node/src/hostClients/whisperMessageBuilder.ts
```diff
@@ -230,6 +230,27 @@ export const generateWhisperListElement = (
       msg.setDivider(inputMsg);
       break;
     }
+    case 'link': {
+      const inputMsg = new WLE.Link();
+      if (element.align) {
+        inputMsg.setAlign(generateWhisperListAlign(element.align));
+      } else {
+        inputMsg.setAlign(generateWhisperListAlign(WhisperListAlign.LEFT));
```
richseviora, line 238: Does it make sense consolidate this into the line above?

something like

```typescript
inputMsg.setAlign(generateWhisperListAlign(element.align || WhisperListAlign.LEFT));
```

richseviora (approved):
Optional change - otherwise looks great!

bbrewer-oliveai (approved):

ldk/node/src/hostClients/whisperMessageBuilder.ts
BrettMugglinOlive: Made that change

PR #55, opened 2021-02-06
Side 1325 node self test loop

BrettMugglinOlive:


ldk/node/examples/selfTestLoop/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "olive-helps-example-clipboard",
+  "name": "node-selft-test-loop",
```
bbrewer-oliveai, line 2:
```suggestion
  "name": "node-self-test-loop",
```

ldk/node/examples/selfTestLoop/package.json
```diff
@@ -29,6 +29,7 @@
   },
   "dependencies": {
     "@oliveai/ldk": "file:../..",
+    "moment": "^2.29.1",
```
bbrewer-oliveai, line 32: It's worth noting that momentjs is "deprecated" and in maintenance mode. They do urge people to adopt a different util for this FYI.

ldk/node/examples/selfTestLoop/src/testingFixtures/testSuite.ts
```diff
@@ -12,12 +17,56 @@ export default class TestSuite {
...
+      markdown: '',
+      elements,
+    });
+    const markdown = host.whisper.markdownWhisper({
+      label: 'Self Test - Results',
+      markdown: 'Press "Ctrl + /" to bring back up the original whisper',
+    });
+    const keyboard = host.keyboard.streamHotKey(hotkeys, (error, response) => {
+      if (error) {
+        // Do nothing
```
bbrewer-oliveai, line 63: Is there a reason we just toss the error?

bbrewer-oliveai (approved):
Looks pretty straightforward though most of the self-test stuff looked to be in place and this is adding on top?

Otherwise, is there a reason this is touching C# and Go? Is it this based off of Rich's branch?

ldk/node/examples/selfTestLoop/src/testingFixtures/testSuite.ts
BrettMugglinOlive: This specific listener is to get back to the starting menu to pick which group you want to test. My thinking here being that if the tests works, this should work too. So why expect an error to be thrown?

ldk/node/examples/selfTestLoop/package.json
BrettMugglinOlive: I think I'm fine keeping it as moment for now, we can address it later

PR #56, opened 2021-02-10
[SIDE-1508] Unblock Close without waiting for stream to close

richseviora:
This PR fixes a bug with the Go LDK preventing the `GRPCFile.Close` function from completing on the client side.

### DIFF
--- a/ldk/go/filesystemClient.go
+++ b/ldk/go/filesystemClient.go
@@ -289,14 +289,10 @@ func (f *GRPCFile) Close() error {
 		},
 	})
 	if err != nil {
+		println("[ERROR] GRPCFile.Close - Received error sending Close", err.Error())
 		return err
 	}
-	select {
-	case <-f.ctx.Done():
-		return nil
-	case <-f.stream.Context().Done():
-		return nil
-	}
+	return nil
 }
 
 // Chown changes owner of file


bbrewer-oliveai (approved):
I like it.

ldk/go/filesystemClient.go
```diff
@@ -289,14 +289,10 @@ func (f *GRPCFile) Close() error {
 		},
 	})
 	if err != nil {
+		println("[ERROR] GRPCFile.Close - Received error sending Close", err.Error())
 		return err
 	}
-	select {
-	case <-f.ctx.Done():
```
DanielWagersOlive, line 295: Do we know why we weren't receiving either of these messages?

richseviora: So ... the [ServerStream](https://pkg.go.dev/google.golang.org/grpc#ServerStream) doesn't give us a mechanism to explicitly cancel the stream; instead the Server returns from the method and that is how the stream is closed.

It doesn't look like the GRPC implementation gives clients in Bidirectional Streaming calls a mechanism to know a call is closed without trying to receive one last message and receiving `io.EOF`.

DanielWagersOlive: Could I theoretically exhaust my available file handles by running `Open` in a loop then?

richseviora: Not the file handles themselves; those are `defer file.Close()` upon exit from the server's loop.

BrettMugglinOlive (approved):

PR #57, opened 2021-02-11
SIDE-1326: Create Go Self-Test Loop

DanielWagersOlive:
In `self-test-loop` there are two source files:
- `main.go` - Enumerates and gives and example for all the functionality of the LDK
- `util/reporter.go` - Runs in the background and reports state to the log every second

Additionally:
- There are listeners registered for every sensor.
- There's a playground whisper that drives triggered events.

In the future, we may report test results via a companion server.


bbrewer-oliveai:
I'd approve if I hadn't helped write it

BrettMugglinOlive (approved):

PR #58, opened 2021-02-12
SIDE-1556 Fix Segmentation / Abort trap race condition from calling steam.stop()

BrettMugglinOlive:
With this change in place on the self test loop branch, everything works.


ldk/node/dist/hostClients/transformingMessage.js
```diff
@@ -21,7 +21,10 @@ class TransformingMessage {
         return this.callbackPromise;
     }
     stop() {
-        this.call.cancel();
+        // SIDE-1556: Needs to be wrapped this way so that we don't trigger a race condition
+        setImmediate(() => {
```
DanielWagersOlive, line 25: So this bumps a function to the front of the event queue?  Neat.  TIL.

DanielWagersOlive (approved):

richseviora (approved):

PR #61, opened 2021-02-23
SIDE-1540 Add "markdown" attribute to disambiguation and list whispers

BrettMugglinOlive:


ldk/go/whisperContentDisambiguation.go
```diff
@@ -4,6 +4,7 @@ import "github.com/open-olive/loop-development-kit/ldk/go/proto"
 
 type WhisperContentDisambiguation struct {
 	Label    string                                         `json:"label"`
+	Markdown string											`json:"markdown"`
```
bbrewer-oliveai, line 7: Pretty sure golint is gonna freak out about this.

bbrewer-oliveai (approved):
Looks good though I'm curious how/if you tested this. Manually with a self-test loop is an acceptable answer.

ldk/go/whisperContentDisambiguation.go
BrettMugglinOlive: Yeah, you're right. I go fmted the code

ldk/csharp/OliveHelpsLDK/OliveHelpsLDK/Whispers/IWhisperService.cs
```diff
@@ -143,6 +143,11 @@ public struct WhisperList
         /// </summary>
         public WhisperConfig Config;
 
+        /// <summary>
+        /// Markdown formatted string displayed at the top of the form.
+        /// </summary>
+        public string Markdown;
```
richseviora, line 149: I think this also requires a corresponding change to `ldk/csharp/OliveHelpsLDK/OliveHelpsLDK/Whispers/WhisperFormBuilder.cs:26`

richseviora (changes requested):
Needs an additional change in .net

ldk/csharp/OliveHelpsLDK/OliveHelpsLDK/Whispers/IWhisperService.cs
BrettMugglinOlive: If this is a change for just WhisperList, would I need to update the form builder?

richseviora: Yes; the `WhisperFormBuilder` also has a `BuildRequest(WhisperList list, Proto.Session session)` method where the message is constructed.

richseviora (approved):

bbrewer-oliveai (approved):

PR #64, opened 2021-02-26
Side-1585 Fix hotkey stream issues

BrettMugglinOlive:


ldk/go/keyboardService.go
```diff
@@ -20,11 +20,12 @@ func (h Hotkey) Match(v Hotkey) bool {
 		return false
 	}
 	// You can't do a direct bitmask as Alt means Altleft or AltRight.  So checking all combinations is the only safe way to actually compare
-	l, r := h.Modifiers, v.Modifiers
-	return (l.AltLeft() == r.AltLeft() || l.Alt() == r.Alt()) && (l.AltRight() == r.AltRight() || l.Alt() == r.Alt()) &&
-		(l.ControlLeft() == r.ControlLeft()) || l.Control() == r.Control() && (l.ControlRight() == r.ControlRight() || l.Control() == r.Control()) &&
-		(l.MetaLeft() == r.MetaLeft() || l.Meta() == r.Meta()) && (l.MetaRight() == r.MetaRight() || l.Meta() == r.Meta()) &&
-		(l.ShiftLeft() == r.ShiftLeft() || l.Shift() == r.Shift()) && (l.ShiftRight() == r.ShiftRight() || l.Shift() == r.Shift())
+	configured, incoming := h.Modifiers, v.Modifiers
```
bbrewer-oliveai, line 23: Not a show stopper, but why is this "almost sensor related logic" in the LDK part of this? Just worth asking as most of our "business logic" is in Olive Helps, not here.

bbrewer-oliveai (approved):

ldk/go/keyboardService.go
BrettMugglinOlive: I think putting this logic here does make sense, since the consumer shouldn't have to be checking what keys were pressed, they just want to know if it's a match

ldk/go/keyboardService.go
```diff
@@ -20,11 +20,12 @@ func (h Hotkey) Match(v Hotkey) bool {
 		return false
 	}
 	// You can't do a direct bitmask as Alt means Altleft or AltRight.  So checking all combinations is the only safe way to actually compare
```
DanielWagersOlive, line 22: Is this true only for Alt?  Or is this true for every left/right modifier?

ldk/go/keyboardService.go
```diff
@@ -20,11 +20,12 @@ func (h Hotkey) Match(v Hotkey) bool {
...
-	l, r := h.Modifiers, v.Modifiers
-	return (l.AltLeft() == r.AltLeft() || l.Alt() == r.Alt()) && (l.AltRight() == r.AltRight() || l.Alt() == r.Alt()) &&
-		(l.ControlLeft() == r.ControlLeft()) || l.Control() == r.Control() && (l.ControlRight() == r.ControlRight() || l.Control() == r.Control()) &&
-		(l.MetaLeft() == r.MetaLeft() || l.Meta() == r.Meta()) && (l.MetaRight() == r.MetaRight() || l.Meta() == r.Meta()) &&
-		(l.ShiftLeft() == r.ShiftLeft() || l.Shift() == r.Shift()) && (l.ShiftRight() == r.ShiftRight() || l.Shift() == r.Shift())
+	configured, incoming := h.Modifiers, v.Modifiers
+
+	return (configured.AltLeft() == incoming.AltLeft() || configured.Alt() == incoming.Alt()) && (configured.AltRight() == incoming.AltRight() || configured.Alt() == incoming.Alt()) &&
```
DanielWagersOlive, line 25: `(a or b) and (a or c)` is equivalent to `a or (b and c)`

Does this seem cleaner?

```
func (h Hotkey) Match(v Hotkey) bool {
	if h.Key != v.Key {
		return false
	}
	// You can't do a direct bitmask as Alt means Altleft or AltRight.  So checking all combinations is the only safe way to actually compare
	hm, vm := h.ModifierMask, v.ModifierMask
	return hm.matchAlt(vm) &&
		hm.matchControl(vm) &&
		hm.matchMeta(vm) &&
		hm.matchShift(vm)
}

func (k KeyModifierBitMask) matchAlt(v KeyModifierBitMask) bool {
	if k.Alt() == v.Alt() {
		return true
	}
	return k.AltLeft() == v.AltLeft() && k.AltRight() == v.AltRight()
}

func (k KeyModifierBitMask) matchControl(v KeyModifierBitMask) bool {
	if k.Control() == v.Control() {
		return true
	}
	return k.ControlLeft() == v.ControlLeft() && k.ControlRight() == v.ControlRight()
}

func (k KeyModifierBitMask) matchMeta(v KeyModifierBitMask) bool {
	if k.Meta() == v.Meta() {
		return true
	}
	return k.MetaLeft() == v.MetaLeft() && k.MetaRight() == v.MetaRight()
}

func (k KeyModifierBitMask) matchShift(v KeyModifierBitMask) bool {
	if k.Shift() == v.Shift() {
		return true
	}
	return k.ShiftLeft() == v.ShiftLeft() && k.ShiftRight() == v.ShiftRight()
}
```

DanielWagersOlive: This is going to be nasty no matter what we do.  I went through several revisions of simplifying to end up here.  Hope it helps.

ldk/go/keyboardService.go
BrettMugglinOlive: I'm not 100% sure what the the bitmask shifting is doing, so I don't know if I can answer that

ldk/go/keyboardService.go
BrettMugglinOlive: It does look nicer than what came before it, however, the hasOneMatchingBitInEachMask function I think needed a little tweaking. "result" needed to start as false and then the line where you update the value would then be result = result || match. Otherwise, it was always outputting false.

DanielWagersOlive (approved):

PR #65, opened 2021-03-01
WIP - consistency

bbrewer-oliveai:
For the moment this is just NodeJS as these are pretty extreme changes to make things consistent with Go and C#.


bbrewer-oliveai:
Ugh, wrong source

PR #66, opened 2021-03-01
WIP - consistency changes

bbrewer-oliveai:
Only NodeJS for now, as the changes are pretty extreme.


richseviora:
Looking through this, I've seen the following changes:

- Remove `Service` from the interface names for the aptitudes, so `BrowserService` becomes `Browser`.
- Rename methods so they're named by the data being returned if we're getting current state, prepended with `listen` if we're setting up an ongoing stream, and other names as appropriate. We're also explicitly removing duplication so `Storage.Read` isn't `Storage.StorageRead` and the like.

I think I'm fine with the changes listed above!

PR #67, opened 2021-03-05
SIDE-1652 Add disambiguation whisper to C#

BrettMugglinOlive:
Shoulda been there


ldk/csharp/OliveHelpsLDK/OliveHelpsLDK/Whispers/Disambiguation.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// Base class for Whisper Disambiguation responses.
+    /// </summary>
+    public interface IWhisperDisambiguationResponse
+    {
+    }
+
+    /// <summary>
+    /// Represents a selection of an item
+    /// </summary>
+    public struct DisambiguationResponse : IWhisperDisambiguationResponse
```
richseviora, line 13: I'm thinking that whenever the namespace includes the relevant noun that it's perhaps redundant to include the noun again in the object name.

So this becomes `OliveHelpsLDK.Whispers.Disambiguation.Response`. 

@RusGraf what do you think?

richseviora: And it's entirely possible I have not been consistent/following this myself :)

ldk/csharp/OliveHelpsLDK/OliveHelpsLDK/Whispers/WhisperClient.cs
```diff
@@ -67,6 +68,27 @@ public Task<bool> ConfirmAsync(WhisperConfirm message, CancellationToken cancell
...
+        {
+            var request = Builder.BuildRequest(message, CreateSession());
+            var call = Client.WhisperDisambiguation(request, new CallOptions(cancellationToken: cancellationToken));
+            var loggedParser = LoggedParser<Proto.WhisperDisambiguationStreamResponse, IWhisperDisambiguationResponse>(response =>
+                Parser.ParseResponse(response)); 
+            return new StreamingCall<WhisperDisambiguationStreamResponse, IWhisperDisambiguationResponse>(call, loggedParser);
+        }
+
+        /* 
+        Need to look into the parser
```
richseviora, line 81: What needs to be checked?

richseviora (approved):
Overall fine - one opportunity to rename some classes but not sure if we should

ldk/csharp/OliveHelpsLDK/OliveHelpsLDK/Whispers/Disambiguation.cs
RusGraf: I would think you are correct if the situation be like:
```
public class WhisperDisambiguation 
{
    public Response {get; set;} // not public WhisperDisambiguationResponse {get; set;}
}
```

But it is totally correct to have Foo : IFoo. To be honest I don't mind long class or struct names. As long as they describing the intent.

RusGraf (changes requested):
I think there is an outstanding question about commented out code. Please respond. Otherwise all good.

ldk/csharp/OliveHelpsLDK/OliveHelpsLDK/Whispers/Disambiguation.cs
BrettMugglinOlive: Ok, I can update those names

ldk/csharp/OliveHelpsLDK/OliveHelpsLDK/Whispers/WhisperClient.cs
BrettMugglinOlive: Nothing, that code block needs to be removed

RusGraf (approved):

PR #69, opened 2021-03-09
SIDE-1511 remove keyboard service.stream scan code

RusGraf:
Removing a scan code service from c#, node and GO implementations. As it is a dead code.


ldk/go/self-test-loop/main.go
```diff
@@ -149,12 +149,7 @@ func (loop *Loop) listenToKeyboard() error {
 		return err
 	}
 
-	err = loop.sidekick.Keyboard().ListenScancode(loop.ctx, loop.keyboardScancodeHandler())
-	if err != nil {
-		loop.logger.Error("keyboard scancode listen error", err)
-	}
-
-	return err
+return err
```
DanielWagersOlive, line 152: nit: missing indent

RusGraf: Updated

ldk/node/package-lock.json
```diff
@@ -1,8 +1,10679 @@
 {
   "name": "@oliveai/ldk",
-  "version": "2.0.0-alpha",
-  "lockfileVersion": 1,
+  "version": "2.0.0-beta",
```
DanielWagersOlive, line 3: What happened here?

DanielWagersOlive (approved):

ldk/node/package-lock.json
RusGraf: Not sure actually, looks like it decided to move a line down.

PR #72, opened 2021-03-10
Fix check annotation issues

DanielWagersOlive:
- Fix some minor issues and remove PR noise
- Delete accidentally committed binary and add a `.gitignore`


DanielWagersOlive:
This is a bigger job than I realized.  We have a lot of small nit-picky issues.  We'll have to fix these as we go.

PR #73, opened 2021-03-17
SIDE-1686: adding go loop documentation

RusGraf:


ldk/go/readme.md
```diff
@@ -11,9 +11,76 @@ go get -u github.com/open-olive/loop-development-kit/ldk/go
...
+		return err
+	}
+
+	if err := sidekick.UI().ListenSearchbar(l.ctx, handler); err != nil {
+		return err
+	}
+}
+```
+
+Loops do not need to emit whispers in a 1:1 relationship with events. Loops may not use events at all. Loops may only use some events. Loops may keep a history of events and only emit whispers when several conditions are met.
```
BrettMugglinOlive, line 61: Beginning each sentence here with "Loops" feels a little...samey? It just doesn't flow very well.

It doesn't have to be exactly this, but something like

"Loops do not need to emit whispers in a 1:1 relationship with events, and may not use events at all. They could only use some events, keep a history of events, or only emit whispers when several conditions are met."

BrettMugglinOlive (approved):

ldk/go/readme.md
```diff
@@ -11,9 +11,76 @@ go get -u github.com/open-olive/loop-development-kit/ldk/go
...
+Writing a loop boils down to writing an implementation for the `Loop` interface and serving it via `ldk.ServeLoopPlugin`.
+
+```go
+type Loop interface {
+	LoopStart(Sidekick) error
+	LoopStop() error
+}
+
+func main() {
+	...
```
richseviora, line 30: in place of the ellipsis, could you show them how we instantiate the logger, and maybe a placeholder for the Loop struct instantiation

ldk/go/readme.md
```diff
@@ -11,9 +11,76 @@ go get -u github.com/open-olive/loop-development-kit/ldk/go
...
-Example loops are available in the examples folder.
+This LDK can be used to write loops for Olive Helps. An example of the loop implementation is available [here](docs/loops.md).
+
+
+#### Interface
+
+Writing a loop boils down to writing an implementation for the `Loop` interface and serving it via `ldk.ServeLoopPlugin`.
+
+```go
+type Loop interface {
```
richseviora, line 24: This reads as a part of the code below, and I don't _think_ it should be? Maybe break it from the `main` function below as a separate codeblock, so it'd be like:
```go
type Loop interface {
  LoopStart(ldk.Sidekick) error
  LoopStop() error
}
```

```go
import $LDK_MODULE$

func main() {
  # Instantiate Loop
  # Instantiate Logger
  ldk.ServeLoopPlugin(logger, loop)
}
```

ldk/go/readme.md
```diff
@@ -11,9 +11,76 @@ go get -u github.com/open-olive/loop-development-kit/ldk/go
...
+
+	if err := sidekick.UI().ListenSearchbar(l.ctx, handler); err != nil {
+		return err
+	}
+}
+```
+
+Loops do not need to emit whispers in a 1:1 relationship with events. Loops may not use events at all. Loops may only use some events. Loops may keep a history of events and only emit whispers when several conditions are met.
+
+#### List of possible Sensors to subscribe to
```
richseviora, line 63: Now that I think about it, and taking to account the developer hub aptitude list we have available, maybe we should link to that instead...

ldk/go/docs/loops.md
```diff
@@ -1,51 +1,108 @@
...
+type Loop struct {
+	ctx              context.Context
+	cancel           context.CancelFunc
+	sidekick         ldk.Sidekick
+	logger           *ldk.Logger
+}
 
-**LoopStop** - The Loop should stop operating when this is called.
+// LoopStart will be called by ldk when loop starts
+func (loop *Loop) LoopStart(sidekick ldk.Sidekick) error {
```
richseviora, line 33: Go's [code style](https://github.com/golang/go/wiki/CodeReviewComments#receiver-names) prefers that receiver variable names are one or two character

richseviora (changes requested):
Some changes :)

ldk/go/readme.md
RusGraf: done

ldk/go/readme.md
RusGraf: done

ldk/go/readme.md
RusGraf: Separated.

ldk/go/readme.md
RusGraf: Not a bad idea. Do you have a link to where would we like to point them?

ldk/go/docs/loops.md
RusGraf: Understood.

ldk/go/readme.md
richseviora: Hmm... not a great link right now - but I like having the interface available instead.

Oops! one thing though, we need to replace references to "Sensors" with "Aptitudes" in the documentation, we don't call them Sensors any more so we should avoid them in the documents.

RusGraf: Ahh, I did that and then changed it back based on the response in the slack. I'll change it again.

RusGraf: Updated

richseviora (approved):

PR #74, opened 2021-03-18
Side 1836 - add window client to nodejs ldk

RusGraf:


BrettMugglinOlive (approved):

ldk/node/.gitignore
```diff
@@ -1,51 +1,166 @@
...
-# Testing
-coverage/
+lerna-debug.log*
+
+# Diagnostic reports (https://nodejs.org/api/report.html)
+report.[0-9]*.[0-9]*.[0-9]*.[0-9]*.json
+
+# Runtime data
+pids
+*.pid
```
richseviora, line 15: Should we be making all these changes here instead of the root `.gitignore`? I think there might be some duplication between some of these entires

richseviora (approved):
Awesome!

ldk/node/.gitignore
RusGraf: Well. This is a standard .getignore, specific to node js. There should be no issues if it is duplicated.

PR #80, opened 2021-03-26
Side 1904 make clipboard sensor trigger configurable

RusGraf:


ldk/csharp/OliveHelpsLDK/OliveHelpsLDK/Clipboard/ClipboardClient.cs
```diff
@@ -42,11 +42,12 @@ public Task Write(string contents, CancellationToken cancellationToken = default
                 .ResponseAsync;
         }
 
-        public IStreamingCall<string> Stream(CancellationToken cancellationToken = default)
+        public IStreamingCall<string> Stream(bool includeOliveHelpTraffic = false, CancellationToken cancellationToken = default)
```
cshields-oliveai, line 45: Should the C# implementation match the Go implementation? How close do these usually match? (Create a `ClipboardListenConfiguration` object)

richseviora: Hmm. Adding a parameter as we do here is a breaking change, and in the future we may decide to add other options, and if we add it as a configuration object now that'd make it easier.

I don't think we have been explicitly keeping the configuration objects identical (as adjusted for language conventions around casing and the like), but maybe that's something we should do.

It's probably a moot point now, but it makes me think that I should've made the `CancellationToken` parameter the first one by convention so that we could add additional parameters without breaking the API.

richseviora (approved):

ldk/csharp/OliveHelpsLDK/OliveHelpsLDK/Clipboard/ClipboardClient.cs
RusGraf: Well, ideally yes. Although C# is not using Sidekick listeners, so it will defer a little bit at the moment. I created a card to provide same sidekick capability for the C# implementation. It is currently on hold due to possibility of dropping Go and C# in the nearest future.

PR #89, opened 2021-04-06
SIDE-1972: RG|CS - adding value field for the select and radio controls

RusGraf:


RusGraf:
> I only see Go changes here; nothing for .NET or Node?

Good catch, fixed.

BrettMugglinOlive (approved):

richseviora:
Still don't see changes made to the Node LDK exposed to the user

richseviora (approved):

PR #91, opened 2021-04-09
Side 1996 sandbox capability ldk setup

cshields-oliveai:


.github/workflows/goja-node.js.yml
```diff
@@ -0,0 +1,37 @@
...
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - run: npm set audit false
+      - run: npm ci
+      - run: npm audit --production --audit-level=low
+      - run: npm run build --if-present
+      - run: npm run lint
+      - run: npm run test
+      # - run: npm run doc
```
cshields-oliveai, line 37: TODO: Add documentation?

ldk/goja/src/index.ts
```diff
@@ -0,0 +1,3 @@
+import * as clipboard from './clipboard';
+
+export { clipboard };
```
cshields-oliveai, line 3: TODO: Clipboard will need comments, and we need clipboard lock, listener cancellation which will be resolved in SIDE-2006.

RusGraf (approved):

.github/workflows/goja-node.js.yml
richseviora: I've been thinking about this; I think we will probably want to generate documentation again with Typedoc to the LDK/go folder in the GH pages repo. There's aspects of the current Typedoc presentation I don't love but we don't need to solve that right now.

ldk/goja/.eslintrc.js
```diff
@@ -0,0 +1,33 @@
...
+    ecmaVersion: 12,
+    sourceType: 'module',
+  },
+  plugins: [
+    '@typescript-eslint',
+  ],
+  rules: {
+    'class-methods-use-this': 'off',
+    'import/no-extraneous-dependencies': ['error', { devDependencies: true }],
+    'import/prefer-default-export': 'off', // Default exports are bad.
```
richseviora, line 22: they really bad

ldk/goja/src/index.ts
richseviora: This was a placeholder, but now I'm thinking that we should expose an instance of the `ClipboardImpl` class.

richseviora (changes requested):
Minor change that we could make but isn't required yet for this initial work

.github/workflows/goja-node.js.yml
cshields-oliveai: I will create a card for this.

cshields-oliveai: https://crosschx.atlassian.net/browse/SIDE-2043

richseviora (approved):

PR #94, opened 2021-04-09
SIDE-2017 Goja UI aptitude

BrettMugglinOlive:


ldk/goja/src/ui/ui.test.ts
```diff
@@ -0,0 +1,54 @@
+import { UIImpl } from '.';
+import { mocked } from 'ts-jest/utils';
+import any = jasmine.any;
+import anything = jasmine.anything;
```
cshields-oliveai, line 4: Are these used anywhere? If they are, I think we should be using the [Jest Any Matcher](https://jestjs.io/docs/expect#expectanyconstructor) instead of jasmine.

BrettMugglinOlive: Yeah, I could do that. I was just following the pattern from the clipboard sensor

ldk/goja/src/ui/index.ts
```diff
@@ -0,0 +1,14 @@
+export interface UI {
+  listenSearchbar(cb: (val: string) => void): void;
+  listenGlobalSearch(cb: (val: string) => void): void;
```
RusGraf, line 3: I think it will be valuable to add comments for the public interfaces.

ldk/goja/src/ui/index.ts
```diff
@@ -0,0 +1,14 @@
...
+}
+
+export class UIImpl implements UI {
```
RusGraf, line 6: Please remove class structure, similar as clipboard aptitude. Based on our discussion.

ldk/goja/src/ui/index.ts
cshields-oliveai: We are adding them as follows:

`/**
     * Returns true if the specified vault entry exists.
     * 
     * @param key - The key for the item to check existence.
     * @returns True if the specified vault entry exists.
     */
    exists(key: string): Promise<boolean>;`

cshields-oliveai (approved):

RusGraf (approved):

PR #95, opened 2021-04-12
Side 2011 add cursor aptitude

RusGraf:


ldk/goja/src/clipboard.ts
```diff
@@ -1,28 +1,47 @@
...
 }
 
 function read(): Promise<string> {
   return new Promise<string>((resolve, reject) => {
     try {
-      oliveHelps.clipboard.read((val: string) => resolve(val));
+      oliveHelps.clipboard.read((clipboardText: string) => resolve(clipboardText));
     } catch (e) {
       reject(e);
       // TODO: add console log
```
cshields-oliveai, line 36: did we want to develop some very basic standard for a console log for these here?

RusGraf: I agree, pushed changes.

cshields-oliveai (approved):

ldk/goja/.eslintrc.js
```diff
@@ -20,6 +20,7 @@ module.exports = {
     'class-methods-use-this': 'off',
     'import/no-extraneous-dependencies': ['error', { devDependencies: true }],
     'import/prefer-default-export': 'off', // Default exports are bad.
+    'no-console': 'off',
```
cshields-oliveai, line 23: It has occurred to me that these console logs will appear in our unit test output as the rejection tests execute. We will need to come up with a more reliable way to do these logs.

ldk/goja/src/clipboard.ts
```diff
@@ -1,30 +1,50 @@
...
+   * @returns A Promise resolving with the current contents of the clipboard.
+   */
   read(): Promise<string>;
-  write(val: string): Promise<void>;
-  listen(cb: (val: string) => void): void;
+
+    /**
+   * Writes the provided text into the clipboard.
+   *
+   * @param text
```
cshields-oliveai, line 14: description of this param?

cshields-oliveai (approved):

ldk/goja/.eslintrc.js
richseviora: I wonder if we can mock the console functions... if so we can do that before each test (in a shared before each or something)

richseviora (approved):

ldk/goja/.eslintrc.js
RusGraf: We could abstract console perhaps...

cshields-oliveai: Could potentially mock console.log() globally like with olivehelps. I could give this a shot with the window aptitude card - if it doesn't pan out I will make a task for us to figure out a nicer way.

PR #100, opened 2021-04-13
SIDE-2053: exposing clipboard configuration for node js

RusGraf:


cshields-oliveai (approved):

BrettMugglinOlive (approved):

ldk/node/src/hostClients/clipboardClient.ts
```diff
@@ -33,9 +33,13 @@ export class ClipboardClient extends BaseClient<ClipboardGRPCClient> implements
   }
 
   streamClipboard(listener: StreamListener<string>): StoppableStream<string> {
-    const request = new messages.ClipboardReadStreamRequest().setSession(
-      this.createSessionMessage(),
-    );
+    return this.streamClipboardConfigurable(listener, false);
+  }
+
+  streamClipboardConfigurable(listener: StreamListener<string>, includeOliveHelpsTraffic: boolean): StoppableStream<string> {
```
richseviora, line 39: I'd rather make this an optional parameter to the initial function, so the signature would be:

```
streamClipboard(listener: StreamListener<string>, includeOliveHelpsTraffic = false): StoppableStream<string>
```
Right now we're mapping each of these functions to an aptitude's capabilities, and adding another funciton breaks that.

richseviora (changes requested):

ldk/node/src/hostClients/clipboardClient.ts
RusGraf: I think overloading for the typescript is not supported: https://www.tutorialsteacher.com/typescript/function-overloading
At least for the same return types. Perhaps I misunderstanding the point.

RusGraf: updated, thanks for the suggestion

richseviora (approved):

PR #102, opened 2021-04-13
SIDE-2016 Network Aptitude

BrettMugglinOlive:


ldk/goja/src/clipboard/clipboard.test.ts
```diff
@@ -1,5 +1,5 @@
 import { mocked } from 'ts-jest/utils';
-import { clipboard } from './clipboard';
```
RusGraf, line 2: I'm not oppose, but could I ask why? I thought previous implementation allow a much cleaner scope importing.

ldk/goja/src/network/index.ts
```diff
@@ -0,0 +1,56 @@
...
+   * import * as util from 'util';
+   * ...
+   * var string = new TextDecoder("utf-8").decode(uint8array);
+   * ```
+   */
+  data: Uint8Array;
+  headers: Record<string, string[]>;
+}
+
+function httpRequest(req: OliveHelps.HTTPRequest): Promise<OliveHelps.HTTPResponse> {
```
RusGraf, line 41: Not sure if it is just preference, but should our method naming convention be different from parameters. For example function request(httpRequest: HTTPRequest). Also since you exporting 'HTTPResponse' as part of the network/index.ts you probably should return it instead of the OliveHelps.HTTPResponse. If our external contract changes we could remap it.

ldk/goja/src/clipboard/clipboard.test.ts
BrettMugglinOlive: This was just how I was taught to do things in JS projects, with each folder having an index file and then exporting anything that needed to be from there. If that's not how we want to structure the project, then that's fine. But I personally like having more folder structure in a project so I know where to look for stuff.

cshields-oliveai (approved):

RusGraf (approved):

ldk/goja/src/network/index.ts
BrettMugglinOlive: You're right, I can change that really to use the locally defined interface

PR #106, opened 2021-04-15
Side 2013 add clipboard lock override to goja

RusGraf:


ldk/goja/src/clipboard/clipboard.test.ts
```diff
@@ -33,9 +34,17 @@ describe('Clipboard', () => {
   });
 
   describe('listen', () => {
+    it('setting clipboard olive helps configuration', () => {
```
cshields-oliveai, line 37: Minor point: I like to have test names read Clipboard -> Listen -> 'sets clipboard olive helps configuration'. What you have reads: Clipboard -> Listen -> 'setting clipboard olive helps configuration'.

cshields-oliveai (approved):

ldk/goja/src/clipboard/clipboard.test.ts
RusGraf: good point, fixed.

richseviora (approved):

PR #110, opened 2021-04-20
SIDE-2063 Whisper aptitude

BrettMugglinOlive:


ldk/goja/src/whisper/index.ts
```diff
@@ -0,0 +1,255 @@
+export enum WhisperComponentType {
```
cshields-oliveai, line 1: This could potentially be a card in the future- add more typedoc comments for these to give users more context as to what these properties do within OliveHelps.

cshields-oliveai (approved):
looks good.

ldk/goja/src/whisper/index.ts
BrettMugglinOlive: We don't currently have documentation with this enums in, but it's a good point

ldk/goja/src/types/oliveHelps/index.d.ts
```diff
@@ -158,32 +157,182 @@ declare namespace OliveHelps {
     all(cb: (whispers: Whisper[]) => void): void;
   }
 
-  interface NewWhisper {
-    label: string;
-    components: Component[];
+  enum Urgency {
+    ERROR = 'error',
```
richseviora, line 161: By convention TS enum names are PascalCase

ldk/goja/src/types/oliveHelps/index.d.ts
```diff
@@ -158,32 +157,182 @@ declare namespace OliveHelps {
...
+  interface TextInput extends Component {
+    label: string;
+    onChange?: (value: string) => void;
+    tooltip?: string;
+    value?: string;
+  }
+
+  interface CollapseBox extends Component {
+    children: Array<
+      | Button
```
richseviora, line 258: It looks like this union type is shared; you can declare a type alias:

```
export type ChildrenElement = Button | Checkbox | Link | ListPair | ...
```

and then consume it like you would any other type:

```
interface CollapseBox extends Component {
...
children: ChildrenElement[];
...
}
```

ldk/goja/src/whisper/index.ts
```diff
@@ -0,0 +1,255 @@
...
+  SUCCESS = 'success',
+  WARNING = 'warning',
+}
+
+export interface WhisperComponent {
+  id?: string;
+  type: string;
+}
+
+export interface Button extends WhisperComponent {
```
richseviora, line 50: Right now this type declaration would support an object like:
```
{
  id: 'button',
  type: 'text',
  label: 'This should be a button but it isn't',
}
```

That'd probably not be great so instead I'd make the `WhisperComponent` a generic type, so it'd look like this:
```
export interface WhisperComponent<T extends WhisperComponentType> {
  id?: string;
  type: T;
}

export interface Button extends WhisperComponent<WhisperComponentType.Button> {
  ...
}
```

richseviora (changes requested):
Changes around typing

richseviora (approved):

PR #111, opened 2021-04-21
[SIDE-2044] Cancellable Listeners

richseviora:
This PR:

- Adds support for cancelling listeners.
- Standardizes the APIs.

Needs to be merged with this PR: https://bitbucket.org/crosschx/sidekick/pull-requests/796/side-2044-listener-cancellation


ldk/javascript/src/promisify.ts
```diff
@@ -0,0 +1,91 @@
...
+  });
+}
+
+export function promisifyWithFourParams<TP1, TP2, TP3, TP4, TOut>(
+  p1: TP1,
+  p2: TP2,
+  p3: TP3,
+  p4: TP4,
+  arg: OliveHelps.ReadableWithFourParams<TP1, TP2, TP3, TP4, TOut>,
+): Promise<TOut> {
```
RusGraf, line 49: Is there a way to have arguments grouped, like: ...args: string[]?

ldk/javascript/src/filesystem/index.ts
```diff
@@ -45,6 +53,25 @@ export enum WriteOperation {
...
+  data: Uint8Array;
+  /**
+   * indicates if file should be overwritten or appended with the provided data
+   */
+  writeOperation: WriteOperation;
+  /**
+   * file mode and permission bits
+   */
+  writeMode: WriteMode;
+}
```
RusGraf, line 73: Thanks for grouping those ;)

ldk/javascript/src/keyboard/keyboard.test.ts
```diff
@@ -20,16 +20,16 @@ describe('Keyboard', () => {
             };
 
             keyboard.listenHotkey(hotkey, callback);
-            expect(oliveHelps.keyboard.listenHotkey).toHaveBeenCalledWith(hotkey, callback);
+            expect(oliveHelps.keyboard.listenHotkey).toHaveBeenCalledWith(hotkey, callback,expect.any(Function));
         });
 
-        it('throws exception when olive helps call fails', () => {
+        it('rejects with the error when the underlying call throws an error', () => {
```
RusGraf, line 26: Wondering if this is still valid after the refactor. All what listenHotkey() doing is calling promisifyListenableWithParam(). Perhaps such test belong in the promisify.test.ts. In this case it will only be tested once and not from every aptitude which depends on this call. (More of the unit test approach.)

RusGraf:
Just a thought. Perhaps we should let loopers know when we merge this pull request. As it is changing the contract.

ldk/javascript/src/promisify.ts
richseviora: Yeah I wasn't a big fan of that function either.

I think we might not be able to do it easily. 

So... we can't just do a straight splat without sacrificing the type safety of the `OliveHelps.ReadableWithFourParams` I don't think. I think it might've been different if we had established the pattern that the callbacks would always be first.

ldk/javascript/src/keyboard/keyboard.test.ts
richseviora: I think it depends on how deeply we want these tests to operate. These tests, after the refactor, are no longer technically unit tests b/c they are also testing the promisify functions. However, keeping them as-is would still validate that the whole chain is interacting with the OH functions as expected.

ldk/javascript/src/promisify.ts
cshields-oliveai: Only way I can think of is some sort of builder pattern: `promisify<T>()` function, which returns itself and then include another function `withParam<TParam>(arg: TParam)` which also returns a reference to that original `promisify<T>()` function. Finally a `.get()` or `.build()` on the end which returns `Promise<TOut>`.

Not sure this makes things more clear or not.

ldk/javascript/src/promisify.ts
```diff
@@ -0,0 +1,91 @@
...
+    try {
+      arg(param, (cb: TOut) => resolve(cb));
+    } catch (e) {
+      console.error(e);
+      reject(e);
+    }
+  });
+}
+
+export function promisifyWithTwoParams<TParam1, TParam2, TOut>(
```
cshields-oliveai, line 28: Some of these are called `<TParamX>` and some are `<TPX>`. I'd be consistent at least in this file.

ldk/javascript/src/filesystem/filesystem.test.ts
```diff
@@ -311,18 +311,25 @@ describe('Filesystem', () => {
...
 
       return expect(actual).rejects.toBe(exception);
     });
   });
 
   describe('join', () => {
     it('returns a promise result with given joined path', () => {
-      const segments = ["first", "second", "third"];
-      const expected = "first/second/third";
+      const segments = ['first', 'second', 'third'];
```
cshields-oliveai, line 327: ' quote all the things

ldk/javascript/src/filesystem/index.ts
```diff
@@ -74,15 +101,15 @@ export interface Filesystem {
    * @param path - path of directory to listen.
    * @param callback - the callback function that's called when a file in the directory changes.
    */
-  listenDir(path: string, callback: (fileEvent: FileEvent) => void): void;
+  listenDir(path: string, callback: (fileEvent: FileEvent) => void): Promise<Cancellable>;
```
cshields-oliveai, line 104: I really do like the way `Promise<Cancellable>` reads.

cshields-oliveai (approved):

ldk/javascript/src/promisify.ts
RusGraf: Yeah, perhaps if we think of something better we could always come back to it.

RusGraf (approved):

PR #112, opened 2021-04-21
Side 2043 documentation

cshields-oliveai:


ldk/goja/typedoc.json
```diff
@@ -0,0 +1,25 @@
...
+    "exclude": [
+      "coverage/**/*",
+      "jest.config.js",
+      "**/*.test.ts"
+    ],
+    "listInvalidSymbolLinks": true,
+    "inputFiles": ["src/index.ts"],
+    "stripInternal": true,
+    "theme": "pages-plugin",
+    "name": "Secure Loop Development Kit for Node.js",
```
RusGraf, line 17: Should it be node.js or something else?

cshields-oliveai: Good question. How about "Javascript Loop Development Kit"?

RusGraf: Sure why not ;)

RusGraf (approved):
image

BrettMugglinOlive (approved):

richseviora:
This doesn't appear to be working. What I think is likely the problem is that both the Goja and Node doc actions are generating files and then overwriting the contents of the `gh-pages` branch. You'll instead want to have one action that builds both doc files and then pushes it via the `actions-gh-pages` action to the one folder.

PR #115, opened 2021-04-22
SIDE-2096: RG - adding version to a webpack banner

RusGraf:


BrettMugglinOlive (approved):

richseviora:
TIL that npm inserts package versions into env vars... do we know if it inserts the actual version, or the sem ver specified?

ldk/goja/src/webpack/generate-banner.ts
```diff
@@ -1,6 +1,9 @@
 export function generateMetadata(): string {
   const json = JSON.stringify({
-    ldkVersion: 'PLACEHOLDER',
+    /**
+     * referencing the ldk dependency version from loop package file, so it could be included as metadata of the minimized loop
+     */
+    ldkVersion: process.env.npm_package_dependencies__oliveai_ldk,
```
richseviora, line 6: OK so this was a good first stab at the solution, but I did some poking around and I think this will run into some problems.

The version provided appears to be the requested version, like so:
```
npm_package_devDependencies_eslint_plugin_import=^2.22.1
```
so that means we don't reliably know which LDK version we're actually getting. That would permit any minor version number and so we could actually be getting `2.50.1` in that scenario (this page explains the syntax https://semver.npmjs.com/).

The other problem is that these package.json environment variables are being dropped from npm v7, so we'd need a fix soon anyway.

The one immediate option that comes to mind is that as part of a post-build script that we replace the placeholder value with the LDK version that we're reading from our `package.json`.

As a longer term solution, I've been wondering if we want to update our build process so that we avoid having a `dist` folder in the compilation. We could adopt something like webpack for the build, and webpack can handle the insertion process with one of its plugins.

richseviora (changes requested):
Changes - plz see comment

ldk/goja/src/webpack/generate-banner.ts
RusGraf: Ahh, great point Rich! I updated the pull request.

RusGraf:
This pull request is obsolete  due to changed versioning directions.

PR #118, opened 2021-04-22
SIDE-2090: update gatsby config to support custom domain

EddieAtOlive:
use root path for assets since we're moving to a custom domain with content at the root


richseviora (approved):

RusGraf (approved):
image

PR #114, opened 2021-04-21
Side 2091 export ldk

cshields-oliveai:

PR #119, opened 2021-04-23
Remove remaining Goja file

dmorrison-olive:
I guess this was left dangling from the big merges recently?


cshields-oliveai:
Thanks @dmorrison-olive - I just came to the same realization that this was leftover. I have this same change pending in https://github.com/open-olive/loop-development-kit/pull/114.

PR #120, opened 2021-04-26
Side 2043b documentation

cshields-oliveai:


.github/workflows/docs.yml
```diff
@@ -28,25 +27,41 @@ jobs:
...
+          cp -r ./javascript/docs typedoc-tmp/javascript
+
       - name: Deploy
         uses: peaceiris/actions-gh-pages@v3
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           publish_branch: gh-pages
-          publish_dir: ./ldk/node/docs
-          destination_dir: ldk/node
+          publish_dir: ./ldk/typedoc-tmp
```
richseviora, line 65: Right now I think our GH pages URLs look like:
`https://open-olive.github.io/loop-development-kit/ldk/node/`

the way I'm reading this and the line below, this will take the contents of `/ldk/typedoc-tmp` and publish to `/ldk`

richseviora (approved):

.github/workflows/docs.yml
cshields-oliveai: I _believe_ it will publish to `/ldk/node` **and** `/ldk/javascript` due to both of those directories living within `/ldk/typedoc-tmp`. I have run this build on this branch and both seemed to have been created (unless some other action run published a javascript doc.

They are both live at the moment: 
https://open-olive.github.io/loop-development-kit/ldk/node/
https://open-olive.github.io/loop-development-kit/ldk/javascript/

I believe this is something we will have to keep our eyes on as the `main` branch action completes.

RusGraf (approved):

PR #122, opened 2021-04-26
SIDE-2062 Add self test loop code to javascript LDK

BrettMugglinOlive:


ldk/javascript/example/self-test-loop/generate-banner.js
```diff
@@ -0,0 +1,18 @@
+function generateMetadata() {
```
RusGraf, line 1: I thought, this banner should be in the javascript ldk and not in the loop.

ldk/javascript/example/self-test-loop/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "loop-goja",
```
RusGraf, line 2: Should we update the name to something close to the self-test-loop.

ldk/javascript/example/self-test-loop/webpack.config.js
```diff
@@ -0,0 +1,69 @@
+const webpack = require("webpack");
```
RusGraf, line 1: I think this is also part of the LDK.

ldk/javascript/example/self-test-loop/generate-banner.js
BrettMugglinOlive: That's been removed at this point

ldk/javascript/examples/self-test-loop/src/selfTest.ts
```diff
@@ -0,0 +1,313 @@
...
+      console.log('Error Streaming', 'error', e.toString());
+    }
+  }
+
+  openTestGroups(): void {
+    let allTests = [] as LoopTest[];
+    // eslint-disable-next-line
+    const clickableElements: any[] = [];
+    // eslint-disable-next-line
+    for (const config in testConfig) {
```
BrettMugglinOlive, line 240: Remove this

cshields-oliveai (approved):
Thanks for the overview. Looks good. Should we create new tasks for continuing to add tests for aptitudes?

RusGraf (approved):

PR #128, opened 2021-04-29
SIDE-2141 - adapting new sidekick contract to properly serialize to Uint8Array

RusGraf:
…to Uint8Array


cshields-oliveai (approved):

ldk/javascript/examples/self-test-loop/src/tests/index.ts
```diff
@@ -248,26 +248,6 @@ export const queryDirectory = (): Promise<boolean> =>
...
-//     for (let i = 0; i < str.length; i+=1) {
-//       const code = str.charCodeAt(i);
-//       bytes = bytes.concat([code]);
-//     }
-
-//     return bytes;
-// }
-
 export const createAndDeleteFile = (): Promise<boolean> =>
   new Promise((resolve, reject) => {
```
richseviora, line 252: I didn't pick up on this before but it might be worth refactoring this to be an `async` function with `await` statements. Under the hood `async` functions return promises, and `await` waits for the promise to resolve.

ldk/javascript/src/promisify.ts
```diff
@@ -1,5 +1,22 @@
 import { Cancellable } from './cancellable';
 
+type Mapper<TIn, TOut> = (param: TIn) => TOut;
+
+export function promisifyWithMapper<TParam, TInternalOut, TExternalOut>(
+  param: TParam,
```
richseviora, line 6: I like this!

ldk/javascript/src/network/index.ts
```diff
@@ -20,9 +22,7 @@ export interface HTTPResponse {
   /**
    * The HTTP response as a byte array. To decode into a UTF-8 string you can:
    * ```
-   * import * as util from 'util';
-   * ...
-   * var string = new TextDecoder("utf-8").decode(uint8array);
+   let decodedText = network.decode(data);
```
richseviora, line 25: I'm wondering whether we should accept multiple types for the body and handle the conversion internally.

So if we say that `data` takes `Uint8Array|string`, we can see if the `data` property is a `string` and if so, encode it as UTF-8 for the user.

richseviora (approved):

ldk/javascript/src/network/index.ts
RusGraf: Yes, that makes sense. I could create a card to prioritise.

PR #129, opened 2021-04-29
SIDE-2121 Set metadata format for permissions

cshields-oliveai:


ldk/javascript/readme.md
```diff
@@ -28,6 +28,43 @@ const merged = merge.merge(ldkConfig.default, {
 module.exports = merged;
 ```
 
+### Loop Permissions
+In order to ensure your Loop is executing in a secure manner, you must declare which network URLs, file system paths, and aptitudes your Loop will use.
+
+Permissions are declared inside of the Loop `package.json` within a `permissions` json object.
```
richseviora, line 34: It's very possible that we might find additional LDK metadata we'd want to put in here. I'm wondering if instead of calling this property `permissions`, that we call it `ldk` and nest the `permissions` object under that.

ldk/javascript/readme.md
```diff
@@ -28,6 +28,43 @@ const merged = merge.merge(ldkConfig.default, {
...
+
+```json
+  "permissions": {
+    "urlPermissions": ["google.com/*"],
+    "filesystemPermissions": ["/tmp"],
+    "aptitudePermissions": ["clipboard"]
```
richseviora, line 40: IMO stylistically, repeating the name property name under child properties is a bit clunky in Node/JS. I don't think this is necessarily the case in .Net.

I'm also wondering whether we want to plan in advance for future aptitude configurations. So instead of listing the aptitudes in this object, we could do something like this:

```yaml
permission:
  network:
    domains:
      - *.google.com
      - yahoo.ca   
  filesystem:
    paths:
      - /tmp/**/*
      - /usr/local/bin
  clipboard: true # Includes text only
```

And then if we decide we want to include clipboard images, we can then say:
```yaml
permission:
  clipboard:
    text: true
    image: true
```

ldk/javascript/readme.md
```diff
@@ -28,6 +28,43 @@ const merged = merge.merge(ldkConfig.default, {
...
 ```
 
+### Loop Permissions
+In order to ensure your Loop is executing in a secure manner, you must declare which network URLs, file system paths, and aptitudes your Loop will use.
+
+Permissions are declared inside of the Loop `package.json` within a `permissions` json object.
+
+```json
+  "permissions": {
+    "urlPermissions": ["google.com/*"],
```
richseviora, line 38: I thought we had agreed to domain with subdomain wildcards?

ldk/javascript/src/webpack/generate-banner.ts
```diff
@@ -1,16 +1,53 @@
-export function generateMetadata(): string {
+import { LoopPermissions } from './loop-permissions';
+
+function isPermissionDefined(permission: string[]): boolean {
```
richseviora, line 3: I think this function name is incorrect; should be `isPermissionUndefined` based on the logic I see below

ldk/javascript/readme.md
richseviora: I think we should also consider which aptitudes are enabled automatically, like Whispers.

ldk/javascript/src/webpack/generate-banner.ts
```diff
@@ -1,16 +1,53 @@
...
+  if (permission.length === 0 || 
+      permission === undefined) {
+    return true;
+  } 
+  return false;
+}
+
+function buildUrlPermissions(permissions: LoopPermissions): string {
+  if (isPermissionDefined(permissions.urlPermissions)) {
```
richseviora, line 12: I don't think we should throw an error if this value is not defined; suppose they just don't want to work with network services?

ldk/javascript/src/webpack/generate-banner.ts
```diff
@@ -1,16 +1,53 @@
...
+
+function isPermissionDefined(permission: string[]): boolean {
+  if (permission.length === 0 || 
+      permission === undefined) {
+    return true;
+  } 
+  return false;
+}
+
+function buildUrlPermissions(permissions: LoopPermissions): string {
```
richseviora, line 11: I'd actually like this to return `string[]`. If we return it as a single string then we're asking the loop metadata processor to split the strings.

ldk/javascript/src/webpack/generate-banner.ts
```diff
@@ -1,16 +1,53 @@
...
+export function generateBanner(loopPackage: LoopPermissions): string {
   return `
-/*
----BEGIN-LOOP-JSON-BASE64---
-${generateMetadata()}
----END-LOOP-JSON-BASE64---
-*/
-`;
-}
+  /*
```
richseviora, line 47: This whitespace was inserted intentionally; otherwise the comment has some whitespace prefixed

richseviora (changes requested):

ldk/javascript/src/webpack/generate-banner.ts
cshields-oliveai: I agree 100% for aptitudes. How do you feel about other permissions @richseviora ?

ldk/javascript/readme.md
cshields-oliveai: sold

ldk/javascript/src/webpack/generate-banner.ts
cshields-oliveai: Good catch.

ldk/javascript/src/webpack/generate-banner.ts
cshields-oliveai: I see.

ldk/javascript/readme.md
cshields-oliveai: @iangallOlive I'd like your input here as well.

iangallOlive: Yep I think this make sense (top-level key is `ldk`, with a child `permissions` key)

ldk/javascript/src/webpack/generate-banner.ts
```diff
@@ -1,16 +1,49 @@
-export function generateMetadata(): string {
+import { LoopPermissions } from './loop-permissions';
+
+function isPermissionUndefined(permission: string[]): boolean {
+  if (permission.length === 0 ||
```
iangallOlive, line 4: One tiny quick thing - I think you'd want `if (permission === undefined || permission.length === 0) { ...`, just because the `.length` access will fail if `permission` is `undefined`, so checking that first should guard it

ldk/javascript/src/webpack/generate-banner.ts
```diff
@@ -1,16 +1,49 @@
...
+function isPermissionUndefined(permission: string[]): boolean {
+  if (permission.length === 0 || 
+      permission === undefined) {
+    return true;
+  } 
+  return false;
+}
+
+function buildUrlPermissions(permissions: LoopPermissions): string {
+  if (isPermissionUndefined(permissions.urlPermissions)) {
```
cshields-oliveai, line 12: I'd like to know everyone's opinions on if we should throw an exception and break the webpack build whenever there are no permissions defined.

ldk/javascript/src/webpack/generate-banner.ts
cshields-oliveai: sweet! Need the team to figure out if this is something we would want from an Loop Author perspective.

ldk/javascript/readme.md
```diff
@@ -28,6 +28,43 @@ const merged = merge.merge(ldkConfig.default, {
 module.exports = merged;
 ```
 
+### Loop Permissions
```
iangallOlive, line 31: One specific thing I think we should consider is that you have to provide a reason for why the loop needs access to some (or all) things. I'm thinking of how Apple does it with apps - if an app wants access to the phone's microphone, the app has to specify why (which will be displayed to the user when the app tries to access the microphone).
So **if** we included this, whether we do it at the aptitude level, or the individual item level (like for each file or URL), the loop reviewers **and** the final users will know why the loop is going to read a specific file or make requests to a certain domain.
Now just like with apps, the dev can totally lie, and there's no proving anything. But it at least gives an opportunity.

ldk/javascript/readme.md
tandrewnichols: What will that look like as JSON in once it's all in the header? We're looking right now for

```
{
   "ldkVersion": "semver", // Soon to be "oliveHelpsContractVersion"
}
```

I think additional top level keys are ideal (from our perspective) here. That is, I wouldn't want `ldk` to also appear in this final output. Just the _value_ of that property. So something like

```
{
  "oliveHelpsContractVersion": "semver",
  "permissions": {
    "urlPermissions": "...",
    "filesystemPermissions": "...", // should it be fileSystemPermisions???
    "aptitudePermissions": "..."
  }
}
```

tandrewnichols: To clarify, there's three reasons for that. 1) `ldk` would make the nesting kinda crazy. You'd need `loop.customData.webpackMetadata.ldk.permissions`. 2) It makes validating it a little trickier. Instead of just "does this object have a, b, and c in it?" we have to ask "does this object have a and b and does b also have c and d?" 3) As you can see we're already grouping all of this as `webpackMetadata` because from our perspective, it's just loop-specific custom data, and that fact that it comes from or belongs to the ldk specifically seems like something the loop library doesn't need to know or deal with. From our point of view, it's _all_ ldk in a way.

ldk/javascript/src/webpack/generate-banner.ts
tandrewnichols: Is it possible to have a loop with no permissions? Seems like it would be possible for _one kind_ of permission to be missing (no need to access remote urls for example), but it also seems like _one of them_ would have to be present or the loop wouldn't do anything.

ldk/javascript/src/webpack/generate-banner.ts
tandrewnichols: I was actually going to ask if there's a guarantee that that object will always be there and have a particular shape. Can we know that permissions will be there and will be an object?

ldk/javascript/readme.md
iangallOlive: I think this totally makes sense, organizing things this way. I think we should be as specific as possible with the keys. So like you did with network `domains`, that is good. But for files, maybe it should be `pathGlobs`. Just in case we ever allow file `pathRegexes`. And same for the chance for network `urlGlobs` or something.

I do wonder if do something like: ```
"permissions": {
  "network": {whatever...},
  "clipboard: {},
}
```
would make sense just to do for enabling aptitudes that don't currently have sub-properties. Just trying to set us up for the long-term and potential for backwards-compatibility, so we don't have to expect a boolean now but an object later. And the way we check for aptitude access is just the existence of the aptitude name in the sub-keys.

File: ldk/javascript/src/webpack/generate-banner.ts [hunk 11]
```diff
@@ -1,16 +1,49 @@
...
+    throw new Error("No permission declaration found for URLs. Add Loop permissions for URLs.");
+  } else {
+    return permissions.urlPermissions.toString();
+  }
+}
+
+function buildFilesystemPermissions(permissions: LoopPermissions): string {
+  if (isPermissionUndefined(permissions.filesystemPermissions)) {
+    throw new Error("No permission declaration found for filesystem. Add Loop mermissions for Filesystem.");
```
iangallOlive, line 21: `mermissions` 😬

ldk/javascript/readme.md
iangallOlive: I think the `ldk` key was simply to make it obvious in the `package.json` that everything inside of it is for "Loop stuff". So I'd agree, and I think it was the intent, that only its **contents** would be used.

ldk/javascript/src/webpack/generate-banner.ts
```diff
@@ -1,16 +1,49 @@
...
+function buildAptitudePermissions(permissions: LoopPermissions): string {
+    return permissions.aptitudePermissions.toString();
+}
+
+export function generateMetadata(loopPackage: LoopPermissions): string {
   const json = JSON.stringify({
     ldkVersion: '0.1.0',
+    urlPermissions: buildUrlPermissions(loopPackage),
+    filesystemPermissions: buildFilesystemPermissions(loopPackage),
+    aptitudePermissions: buildAptitudePermissions(loopPackage)
```
tandrewnichols, line 36: Ah, ok, this answers my previous comment, and this seems perfectly fine to me.

ldk/javascript/readme.md
tandrewnichols: Yeah, this makes sense to me.

ldk/javascript/src/webpack/generate-banner.ts
tandrewnichols: We should leave it ;)

ldk/javascript/readme.md
cshields-oliveai: Love this idea.

ldk/javascript/readme.md
cshields-oliveai: We can use `ldk` at the top level, and send metadata withough wrapping it within `ldk` if that works for everyone. I do like the idea of separation within `package.json`.

ldk/javascript/src/webpack/generate-banner.ts
cshields-oliveai: lol

ldk/javascript/src/webpack/generate-banner.ts
cshields-oliveai: I think it may be best to actually **stop** webpack compilation (and return an exception to the Loop Author) when there are no permissions defined.

cshields-oliveai:
I will be updating this PR to abide by the metadata structure:

```
{
    "ldkVersion": string,
    "permissions": {
        "clipboard": {},
        "cursor": {},
        "filesystem": {
            "pathGlobs": [
                { "value": "/some/path/something.txt", "reason": "To read the shared file" }
            ]
        },
        "keyboard": {},
        "network": {
            "urlDomains": [
                { "value": "*.google.com", "reason": "For Google searches" }
            ]
        }
        "process": {},
        "ui": {},
        "vault": {},
        "whisper": {},
        "window": {}
    }
}
```

ldk/javascript/src/webpack/generate-banner.ts
```diff
@@ -1,16 +1,33 @@
...
   return `
 /*
 ---BEGIN-LOOP-JSON-BASE64---
-${generateMetadata()}
+${generateMetadata(ldkSettings)}
 ---END-LOOP-JSON-BASE64---
 */
 `;
-}
+}
```
cshields-oliveai, line 33: Apparently I aggressively hunt newlines at end of files and delete them... I think this is still OK given the formatting we need in the final metadata header.

cshields-oliveai:
After further discussion - we have decided to **not include** `reason` attributes at this time. We may add them at a later date. Due to this, I will be updating this PR again with this change.

epiccoleman (approved):

ldk/javascript/src/webpack/generate-banner.test.ts
```diff
@@ -0,0 +1,57 @@
...
+                    vault: {},
+                    whisper: {},
+                    window: {},
+                }
+            }
+        }
+        const result = getLoopMetadataContent(generateBanner(ldkSettings));
+
+        //* aptitudes *//
+        expect(result).toContain("\"clipboard\":{}");
```
richseviora, line 35: Wondering whether it makes sense to test for string presence as opposed to deserialize JSON and testing deep object equality. This test structure would pass this data structure:
```yaml
ldk:
  permissions:
    unexpectedProperty: 
      clipboard: {}
```

ldk/javascript/src/webpack/generate-banner.ts
```diff
@@ -1,16 +1,33 @@
-export function generateMetadata(): string {
+import { LdkSettings } from './ldk-settings';
+
+export function generateMetadata(ldkSettings: LdkSettings): string {
+  if(Object.keys(ldkSettings.ldk).length === 0) {
+    throw new Error("Please provide LDK settings in your Loop package.json. See README for more information.")
```
richseviora, line 5: We could probably make this error message a bit more explicit about the expected remedy, maybe something like:

> Please add a "ldk" object to your package.json file with a permission property:
>
> ```
> "ldk": {
>   "permissions": {}
> }
> ```

ldk/javascript/src/webpack/generate-banner.ts
```diff
@@ -1,16 +1,33 @@
...
+  }
   const json = JSON.stringify({
     ldkVersion: '0.1.0',
+    permissions: {
+      clipboard: ldkSettings.ldk.permissions.clipboard || undefined,
```
richseviora, line 10: The specification we discussed was that a missing key would be treated as a no-access request... and `JSON.stringify({value: undefined})` returns `"{}"`.

Maybe we should capture that expectation in a comment so we memorialize the expected behaviour.

Actually, maybe we should capture that in a test.

richseviora (approved):
I have some changes I'd like to see, but knowing that this PR has been outstanding for a while I'm fine merging in as-is and making the other changes in a separate PR.

ldk/javascript/src/webpack/generate-banner.ts
cshields-oliveai: `JSON.stringify` will omit properties with `undefined` values, so it's yielding the behavior we expect here. Maybe not the most straightforward way to do it, but we can always iterate on this.

Capturing this behavior in a test is a good idea.

ldk/javascript/src/webpack/generate-banner.test.ts
cshields-oliveai: I like this idea. Quick enough to change it.

ldk/javascript/src/webpack/generate-banner.ts
cshields-oliveai: I accept this as well - we want to make sure folks know what to do when their transpilation fails.

iangallOlive:
One quick thing I realized - someone **just** added a new aptitude, "Environment", for env var access. It might make sense to spearhead that in this PR and add support for an `environment` key like that looks like this:

```
"environment": {
  "names": [{
    "value": "some env var name"
  }, {
    "value": "some other env var name"
  }]
}
```

cshields-oliveai:
Thanks for the reminder there @iangallOlive ! I think our best bet is to get this out there, and create a new PR (and card) to add this `environment` block. I don't believe it would be a breaking change to add that within the `permissions` block.

iangallOlive:
Sounds good! Just wanted to call it out at least

PR #130, opened 2021-04-29
SIDE-2061: Add Environment Aptitude

DanielWagersOlive:
- Modified self-test-loop to include environment tests


RusGraf (approved):
Looks pretty straightforward.

richseviora (approved):

richseviora (changes requested):
Sorry - just realized that since we're adding an aptitude we should also be updating the developer hub documentation and adding an entry for the aptitude and capability

ldk/javascript/src/environment/index.ts
```diff
@@ -0,0 +1,23 @@
...
+     * Returns the value for the specified environment variable.
+     *
+     * @param key - The name of the variable to read.
+     * @returns Value for the specified environment variable.
+     */
+    read(key: string): Promise<string>;
+}
+
+export function read(key: string): Promise<string> {
+    return new Promise<string>((resolve, reject) => {
```
richseviora, line 15: Should be updated to use the promisify functions I've introduced in my new merged PR

DanielWagersOlive: We now have a readable with param and *two* outputs.  This is a unique case.  Do we *really* need to add a generic function wrapper around this?  It would be called something like `ReadableWithParamAndTwoOuts` or something.  Seems kind of excessive and ridiculous.

DanielWagersOlive: Actually, it would end up being `promisifyWithParamAndTwoOutsAndWrappedReturnType`, like this:

```typescript
export function read(name: string): Promise<EnvironmentVariable> {
    return promisifyWithParamAndTwoOutsAndWrappedReturnType(name, oliveHelps.environment.read);
}
```

```typescript
export function promisifyWithParamAndTwoOutsAndWrappedReturnType<TReturn, TParam, TOut, TOut2>(
    param: TParam,
    arg: OliveHelps.ReadableWithParamAndTwoOuts<TParam, TOut, TOut2>,
): Promise<TReturn> {
  return new Promise((resolve, reject) => {
    try {
      arg(param, (val1: TOut, val2: TOut2) => resolve(<TReturn>{val1, val2}));
    } catch (e) {
      console.error(e);
      reject(e);
    }
  });
}
```

richseviora: Hmm... how about we just return two values in the one object? We've thus far kept to returning one result via callback; I don't know if we should be returning multiple objects unless there's some hugely pressing need for it.

PR #133, opened 2021-04-30
[SIDE-2123] Fix Runtime Polyfills

richseviora:
This PR does a number of things:

- Updates the way that we polyfill modern capabilities into the environment so that the user doesn't need to add incrementally new polyfills (like `Symbol.asyncIterator`) as they add more features. This comes at the cost of an increased compilation size.
- Removes the `core-js` entries from the config.
- Uses Babel to transpile TS too.

### DIFF
--- a/ldk/javascript/src/index.ts
+++ b/ldk/javascript/src/index.ts
@@ -1,3 +1,4 @@
+import "core-js";
 import * as clipboard from './clipboard';
 import * as cursor from './cursor';
 import * as filesystem from './filesystem';
--- a/ldk/javascript/src/webpack/config.ts
+++ b/ldk/javascript/src/webpack/config.ts
@@ -45,7 +45,7 @@ const config: webpack.Configuration = {
                 [
                   '@babel/preset-env',
                   {
-                    useBuiltIns: 'usage',
+                    useBuiltIns: 'entry',
                     corejs: '2.6',
                   },
                 ],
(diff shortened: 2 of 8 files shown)


ldk/javascript/src/webpack/config.ts
```diff
@@ -37,7 +36,25 @@ const config: webpack.Configuration = {
...
+              presets: [
+                [
+                  '@babel/preset-env',
+                  {
+                    useBuiltIns: 'entry',
+                    corejs: '3.11',
+                  },
+                ],
+              ],
+              plugins: ['@babel/plugin-transform-destructuring', '@babel/plugin-transform-runtime'],
```
cshields-oliveai, line 53: Curious: why was`@babel/plugin-transform-destructuring` plugin selected. I don't see us using this feature anywhere inside of the `/javascript` project. Is it required by a 3rd party npm module?

cshields-oliveai (approved):

ldk/javascript/src/index.ts
```diff
@@ -1,3 +1,4 @@
+import "core-js";
```
RusGraf, line 1: How is this being used if we are not exporting it?

RusGraf (approved):

ldk/javascript/src/webpack/config.ts
richseviora: So destructing support lets us do this: https://babeljs.io/docs/en/babel-plugin-transform-destructuring. It's not something we're consuming inside the LDK itself I don't think, but our LDK users will often do that.

There's a further question about how we let consumers add additional plugins.

ldk/javascript/src/index.ts
richseviora: This one is polyfilling and modifying the global scope

PR #153, opened 2021-05-05
SIDE-2160: Whisper Return Object in all Component Callbacks 

scottkipferOlive:
Passing the whisper object back with all component callbacks will allow loop authors to utilize the `close` and eventually the `update` methods inside of the components callbacks.

Note: This PR also removes the `all` method from the whisper service.  The `all` method was left over from the thinking that whispers would have to be persisted through a loop shut down. 

- Also add an error parameter to the close method on the whisper object

### DIFF
--- a/ldk/javascript/src/types/oliveHelps/index.d.ts
+++ b/ldk/javascript/src/types/oliveHelps/index.d.ts
@@ -258,6 +258,11 @@ declare namespace OliveHelps {
     Right = 'right',
   }
 
+  interface Whisper {
+    id: string;
+    close(cb: (err: string) => void): void;
+    // update(whisper: NewWhisper, cb: (err: string) => void): void // TODO: Implement 
+  }
   interface Component<T extends WhisperComponentType> {
     id?: string;
     type: T;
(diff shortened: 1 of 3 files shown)


ldk/javascript/src/types/oliveHelps/index.d.ts
```diff
@@ -268,28 +271,28 @@ declare namespace OliveHelps {
   type Button = Component<WhisperComponentType.Button> & {
     buttonStyle?: ButtonStyle;
     label: string;
-    onClick: () => void;
+    onClick: (whisper: Whisper) => void;
```
richseviora, line 274: Starting to think that we should introduce a new type for these Whisper onEvent handlers... not for immediate action right now, just thinking out loud.

richseviora (changes requested):
I've merged in some changes to the documentation, including updating the Whisper aptitude pages (in `/docs/src/markdown-pages/aptitudes/` IIRC). That includes adding the `whisper.all` capability. Could you please drop that file in this PR as well.

richseviora (approved):
Apparently I was having a senior moment; I thought I created the file but I didn't! All good then.

cshields-oliveai (approved):

PR #172, opened 2021-05-11
SIDE-2239: RG - adding permissions to the self test loop

RusGraf:


ldk/javascript/examples/self-test-loop/package.json
```diff
@@ -9,7 +9,42 @@
   "keywords": [],
   "author": "",
   "ldk": {
-    "permissions": {}
+    "permissions": {
```
cshields-oliveai, line 12: Can you format this file?

cshields-oliveai (approved):

richseviora (approved):

PR #173, opened 2021-05-11
added images for vs code extension

Olive-delight:


Olive-delight:
The images are for the VS Code extension readme

Issue #163, opened 2021-05-07
Behavior of keyboard hotkey


hans-hsieh-asus:
When trying out the `keyboard.listenHotkey()` in the javascript LDK, some behaviors look a little strange. I'm not sure if they are bugs or intended behaviors.
- On Mac, it looks like the `meta` key maps to `options` key on the keyboard, and `alt` key maps to the `cmd` key on the keyboard. Shouldn't they be reversed? I'm not that familiar with Mac. Maybe I'm wrong.
- When I press Ctrl + I, all the listeners for Ctrl + I, Ctrl + shift + I, Ctrl + meta + I, and Ctrl + Alt + I will be called. Is that the intended behavior? If so, how do I distinguish the user is pressing Ctrl + I instead of Ctrl + Alt + I?
- If I press Ctrl + I for a few seconds and release, the listener will be called with argument `pressed` being `true` for several times. But it is never called with `false`.

hans-hsieh-asus:
Thank you for your reply. For the first one, because I don't have Mac, I asked my colleagues to test it for me. I'll confirm with her. :)

cshields-oliveai:
Hello @hans-hsieh-asus - I have been investigating point #3:

`If I press Ctrl + I for a few seconds and release, the listener will be called with argument pressed being true for several times. But it is never called with false.`

I am not seeing the same behavior. In my testing, the callback for `keyboard.listenHotkey()` is only called a single time - on first press of the hotkey and not on held hotkeys.

Do you have any more information about your testing that can help us investigate this one? A few things that can help:

- Olive Helps version
- LDK version
- Operating system
- Code examples with your usage of `keyboard.listenHotkey()`

hans-hsieh-asus:
> Hello @hans-hsieh-asus - I have been investigating point 3:
> 
> `If I press Ctrl + I for a few seconds and release, the listener will be called with argument pressed being true for several times. But it is never called with false.`
> 
> I am not seeing the same behavior. In my testing, the callback for `keyboard.listenHotkey()` is only called a single time - on first press of the hotkey and not on held hotkeys.
> 
> Do you have any more information about your testing that can help us investigate this one? A few things that can help:
> 
> * Olive Helps version
> * LDK version
> * Operating system
> * Code examples with your usage of `keyboard.listenHotkey()`

@cshields-oliveai 
Here's the code example:
```javascript
keyboard.listenHotkey({
  key: 'i',
  control: true,
}, (pressed) => {
[...]

PR #175, opened 2021-05-12
[SIDE-2236] Fix hotkey matching logic

richseviora:
This PR:

- Reworks the Hotkey matching logic so that it prevents matches when the actual keypress is a superset of the requested keypress.
- Adds some tests to validate the above.

### DIFF
--- a/ldk/go/keyboardService_test.go
+++ b/ldk/go/keyboardService_test.go
@@ -74,24 +74,25 @@ func TestHotKeyMatch(t *testing.T) {
 		name      string
 	}{
 		{
-			actual:    ldk.Hotkey{Key: 'a', Modifiers: ldk.KeyModifierCommandAlt},
-			requested: ldk.Hotkey{Key: 'a', Modifiers: ldk.KeyModifierCommandAlt},
-			expected:  true,
-		},
-		{
 			name:      "alt vs alt+shift",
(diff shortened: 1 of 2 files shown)


ldk/go/keyboardService_test.go
```diff
@@ -50,7 +51,7 @@ func TestKeyModifiers(t *testing.T) {
 				for j, fn := range funcs {
 					fnName := runtime.FuncForPC(reflect.ValueOf(fn).Pointer()).Name()
 					if i == j {
-						t.Logf("testing %d -> %s", bit, fnName)
+						// t.Logf("testing %d -> %s", bit, fnName)
```
cshields-oliveai, line 54: Is the log needed?

cshields-oliveai (approved):
Looks Good. Reminder - here is the associated Github Issue: https://github.com/open-olive/loop-development-kit/issues/163

ldk/go/keyboardService_test.go
```diff
@@ -64,3 +65,73 @@ func TestKeyModifiers(t *testing.T) {
 		})
 	}
 }
+
+func TestHotKeyMatch(t *testing.T) {
```
cshields-oliveai, line 69: Thank you for adding these.

RusGraf (approved):

ldk/go/keyboardService_test.go
richseviora: Honestly I think I'd rather keep it commented out; it makes the test output really noisy.

PR #177, opened 2021-05-13
Side 2129 add websocket endpoint to network

RusGraf:


ldk/javascript/examples/self-test-loop/src/tests/index.ts
```diff
@@ -751,6 +753,85 @@ export const networkHTTP = (): Promise<boolean> =>
...
+export const networkWebSocket = (): Promise<boolean> =>
+  new Promise((resolve, reject) => {
+    const url = 'ws://html5rocks.websocket.org/echo';
+    const testData = new Uint8Array([53, 6, 6, 65, 20, 74, 65, 78, 74]);
+    const testText = 'some text';
+    let textTestPassed = false;
+    let binaryTestPassed = false;
+
+    setTimeout(() => {
+      reject(new Error('Network websocket test didnt finished in the appropriate timespan.'));
```
cshields-oliveai, line 765: 'Network websocket test did not finish in the appropriate timespan.'

ldk/javascript/examples/self-test-loop/src/tests/index.ts
```diff
@@ -751,6 +753,85 @@ export const networkHTTP = (): Promise<boolean> =>
...
+                setTimeout(() => {                  
+                  // waiting 1 sec to allow callback to finish
+                  resolve(true);
+                }, 1000);
+              }
+            }
+          },
+          onConnectError: (error: Error) => {
+            if (error != null) {
+              console.debug(`Received on connect error: ${JSON.stringify(error)}`);
```
cshields-oliveai, line 809: should we `reject()` on `onConnectError` and on `onDisconnected`?

ldk/javascript/examples/self-test-loop/src/tests/index.ts
```diff
@@ -11,7 +11,8 @@ import {
   filesystem,
 } from '@oliveai/ldk';
 
-import { Cancellable } from '@oliveai/ldk/dist/cancellable'
+import { Cancellable } from '@oliveai/ldk/dist/cancellable';
+import { collapseTextChangeRangesAcrossMultipleVersions } from 'typescript';
```
cshields-oliveai, line 15: What is this?

ldk/javascript/examples/self-test-loop/src/tests/index.ts
```diff
@@ -814,28 +827,108 @@ export const networkHTTP = (): Promise<boolean> =>
       });
   });
 
+const finaliseWebsocketTest = async (cancellable: Cancellable, socket: network.Socket) => {
```
cshields-oliveai, line 830: "finalizeWebsocketTest"

cshields-oliveai: This is a good candidate for a `utilities` file or something. I get why you added it here but it feels a bit out of place.

ldk/javascript/examples/self-test-loop/src/tests/index.ts
```diff
@@ -814,28 +827,108 @@ export const networkHTTP = (): Promise<boolean> =>
...
+      reject(new Error('Network websocket test did not finish in the appropriate timespan.'));
+    }, 20000);
+
+    const socketConfiguration: network.SocketConfiguration = {
+      url: url,
+      useCompression: true,
+    };
+
+    try {
+      const socket = await network.webSocketConnect(socketConfiguration);
```
cshields-oliveai, line 859: How did you get around our linter settings with mixing `async`/`await` and `.then()`?

ldk/javascript/examples/self-test-loop/src/tests/index.ts
```diff
@@ -814,28 +827,108 @@ export const networkHTTP = (): Promise<boolean> =>
...
+      });
+      // send binary
+      await socket.writeMessage(testData, (error) => {
+        if (error) {
+          console.error(`Received on write binary message error: ${error.message}`);
+          reject(error);
+        }
+      });
+    } catch (error) {
+      console.error(`Error received while testing websocket: ${error.message}`);
```
cshields-oliveai, line 901: Do you need to disconnect from socket if you catch this error?

ldk/javascript/src/network/index.ts
```diff
@@ -56,24 +106,25 @@ export interface Network {
...
-}
 
-const mapToHttpResponse = (response: OliveHelps.HTTPResponse) => ({
-  statusCode: response.statusCode,
-  body: new Uint8Array(response.body),
-  headers: response.headers,
-});
+  /**
+   *
+   * @param socketConfiguration a configuration object defines websocket
```
cshields-oliveai, line 112: I would almost add some kind of extra documentation here that calls out "only one listener can be added to each socket" as well.

ldk/javascript/src/network/index.ts
```diff
@@ -29,6 +28,57 @@ export interface HTTPResponse {
...
+   * @param callback function to call if error occured
+   */
+  writeMessage(message: string | Uint8Array, callback: CallbackError): Promise<void>;
+  /**
+   * closes websocket
+   * @param callback function to call if error occured
+   */
+  close(callback: CallbackError): Promise<void>;
+  /**
+   * allows to listen for a websocket message (there should be only 1 listen message callback per socket to allow messages being fully received)
```
cshields-oliveai, line 74: I would update

"there should be only 1 listen message callback per socket to allow messages being fully received"

to

"there **must** be only one listener registered per socket for messages to be fully received"

ldk/javascript/src/utils/mapper.ts
```diff
@@ -0,0 +1,68 @@
...
+  console.error(`Received error calling ${type}: ${error.message}`);
+  reject(error);
+};
+
+enum MessageType {
+  text = 1,
+  binary = 2,
+}
+
+const mapToBinaryData = (message: string | Uint8Array) =>
```
cshields-oliveai, line 23: Can you specify return types for each of these functions?

ldk/javascript/src/network/index.ts
RusGraf: The comment is added on the listenMessage method, listenMessage method is not required to be used which makes this information being specific to a listenMessage method:
`  /**
   * allows to listen for a websocket message (there should be only 1 listen message callback per socket to allow messages being fully received)
   * @param callback receives text or data message from websocket and error if occurs
   */
  listenMessage: (`

ldk/javascript/src/network/index.ts
RusGraf: "In the end, there can be only one." - Accepted!

ldk/javascript/src/utils/mapper.ts
RusGraf: done

ldk/javascript/src/network/index.ts
RusGraf: done

ldk/javascript/examples/self-test-loop/src/tests/index.ts
RusGraf: updated

ldk/javascript/examples/self-test-loop/src/tests/index.ts
RusGraf: Removed, I have no idea where this thing came from. Perhaps a merge, not sure.

cshields-oliveai (approved):

ldk/javascript/src/network/index.ts
```diff
@@ -29,6 +28,64 @@ export interface HTTPResponse {
...
+   * writes message to a websocket
+   * @param message text or data message
+   * @param callback function to call if error occurred
+   */
+  writeMessage(message: string | Uint8Array, callback: CallbackError): Promise<void>;
+  /**
+   * closes websocket
+   * @param callback function to call if error occurred
+   */
+  close(callback: CallbackError): Promise<void>;
```
richseviora, line 72: Can we get the error to reject the promise instead?

ldk/javascript/src/network/index.ts
```diff
@@ -29,6 +28,64 @@ export interface HTTPResponse {
...
+/**
+ * Object to communicate with the websocket
+ */
+export interface Socket {
+  /**
+   * writes message to a websocket
+   * @param message text or data message
+   * @param callback function to call if error occurred
+   */
+  writeMessage(message: string | Uint8Array, callback: CallbackError): Promise<void>;
```
richseviora, line 67: Same thing here; can we get the error to reject the promise ?

ldk/javascript/src/network/index.ts
```diff
@@ -29,6 +28,64 @@ export interface HTTPResponse {
...
+   */
+  listenMessage: (
+    callback: (error: Error | undefined, message: string | Uint8Array) => void,
+  ) => Promise<Cancellable>;
+  /**
+   * allows to provide handler when websocket closing
+   * @param callback receives code status and text received from the peer
+   */
+  onCloseHandler(
+    callback: (error: Error | undefined, code: number, text: string) => void,
```
richseviora, line 85: I wonder if this sort of thing might be better set as a property on the object. So this would be "setOnClose" or maybe just an onClose property that can have a handler assigned.

ldk/javascript/src/network/index.ts
```diff
@@ -29,6 +28,64 @@ export interface HTTPResponse {
...
+   * @param callback function to call if error occurred
+   */
+  writeMessage(message: string | Uint8Array, callback: CallbackError): Promise<void>;
+  /**
+   * closes websocket
+   * @param callback function to call if error occurred
+   */
+  close(callback: CallbackError): Promise<void>;
+  /**
+   * allows to listen for a websocket message (there must be only one listener registered per socket for messages to be fully received)
```
richseviora, line 74: I think we might want to enforce this ourselves instead of telling the user they can't do it.

richseviora (changes requested):

ldk/javascript/src/network/index.ts
RusGraf: It is enforced in the sidekick. It will return error if you try to supply 2 listeners per 1 socket.

ldk/javascript/src/network/index.ts
RusGraf: Done

ldk/javascript/src/network/index.ts
RusGraf: done

ldk/javascript/examples/self-test-loop/src/tests/index.ts
```diff
@@ -921,6 +922,75 @@ export const networkHTTP = (): Promise<boolean> =>
...
+          return;
+        }
+
+        console.info(`Received on close code: ${code}. ${text}`);
+      });
+      const cancellable: Cancellable = await socket.setMessageHandler(async (error, message) => {
+        if (error) {
+          console.error(error);
+
+          return;
```
BrettMugglinOlive, line 960: Probably want to do a `reject()` here, right? Unless you expect that there could be errors from the connection.

ldk/javascript/examples/self-test-loop/src/tests/index.ts
```diff
@@ -921,6 +922,75 @@ export const networkHTTP = (): Promise<boolean> =>
...
+              if (binaryTestPassed) {
+                resolve(true);
+                await testUtils.finalizeWebsocketTest(cancellable, socket);
+              }
+            }
+          } else {
+            if (JSON.stringify(message) === JSON.stringify(testData)) {
+              console.debug(`Received binary data`);
+              binaryTestPassed = true;
+              if (textTestPassed) {
```
BrettMugglinOlive, line 976: What is the purpose of having this and `binaryTestPassed`? Does the connection you're using work in some way where it sends both messages?

ldk/javascript/src/network/index.ts
```diff
@@ -29,6 +28,62 @@ export interface HTTPResponse {
   headers: Record<string, string[]>;
 }
 
+/**
+ * a simplified representation of a callback which take error
```
BrettMugglinOlive, line 32: For all of these comments above your interface, I think we should use sentence case ("This is sentence case")

DanielWagersOlive (approved):

ldk/javascript/src/utils/mapper.ts
```diff
@@ -0,0 +1,94 @@
...
+
+const mapToResponseMessage = (messageType: MessageType, buffer: ArrayBuffer): string | Uint8Array => {
+  const data = mapToUint8Array(buffer);
+  return messageType === MessageType.text ? new TextDecoder().decode(data) : data;
+};
+
+const mapToMessageType = (message: string | Uint8Array): MessageType =>
+  typeof message === 'string' ? MessageType.text : MessageType.binary;
+
+export const mapToSocket = (socket: OliveHelps.Socket): Socket => ({
```
BrettMugglinOlive, line 34: These are all pretty straight forward functions except for this one. I think we should add some documentation about what this function is doing.

BrettMugglinOlive (approved):

ldk/javascript/src/network/index.ts
RusGraf: Updated

PR #191, opened 2021-05-18
SIDE-2273: Adding self-test for directory exists on the filesystem

cshields-oliveai:
We noticed that directory exists was not working properly a few days ago, in my investigation, I added a new self-test around the existence of a directory to investigate.

Upon updating to the latest sidekick, I see that a fix had already been made to the production code there. (versions of OH > OH > v0.17.3)

I have a testing counterpart PR opened with sidekick that can be reviewed separately.

### DIFF
--- a/ldk/javascript/examples/self-test-loop/package.json
+++ b/ldk/javascript/examples/self-test-loop/package.json
@@ -35,6 +35,9 @@
           },
           {
             "value": "./test.txt"
+          },
+          {
+            "value": "./test-tmp-dir"
           }
         ]
       },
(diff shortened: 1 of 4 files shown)


richseviora (approved):

RusGraf (approved):

RusGraf:
Just a thought. Is there could be a value to also use exist function on file? Wondering if there could be potentially a broken exist(file) but exist(dir) still passing.

cshields-oliveai:
Added file exists test per @RusGraf s request.

PR #194, opened 2021-05-20
SIDE-1914: Add User Aptitude

DanielWagersOlive:
- Test in Self Test Loop just checks for a populated token
- Some formatting changes were included


ldk/javascript/src/user/index.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * The User aptitude gives access to Olive Helps user related information
+ */
+import { promisify } from "../promisify";
+
+export interface User {
+    /**
+     * Returns a JWT with the current username in the subject field
```
cshields-oliveai, line 8: You'll want to use the `@returns` annotation here so its parsed by typedoc I do believe.

ldk/javascript/src/user/index.ts
```diff
@@ -0,0 +1,15 @@
+/**
```
cshields-oliveai, line 1: We added basic unit tests for all of the other aptitudes. I'd vote to keep this aptitude consistent (even though they are just asserting that calls are being made to our `oliveHelps` reference).

cshields-oliveai (approved):

ldk/javascript/src/user/index.ts
DanielWagersOlive: Not a fan.  The tests are extremely low value (imo), don't document anything, and probably add more confusion and build time.  But I'll do it if the team values consistency.

ldk/javascript/src/user/index.ts
DanielWagersOlive: Thanks.  Will do.

DanielWagersOlive: Done.

BrettMugglinOlive (approved):

PR #196, opened 2021-05-26
SIDE-2191 - expose update function/ make onClose optional

scottkipferOlive:
This PR:
-  exposes the `whisper.update` function
- make `onClose` optional when creating a whisper.


cshields-oliveai (approved):
Only recommendation I could make here is to add a new test to self-test-loop that includes update.

ldk/javascript/src/types/oliveHelps/index.d.ts
```diff
@@ -290,7 +290,7 @@ declare namespace OliveHelps {
   interface Whisper {
     id: string;
     close: Readable<undefined>;
-    // update(whisper: NewWhisper, cb: (err: string) => void): void // TODO: Implement
+    update(whisper: UpdateWhisper, cb?: (err: Error) => void): void
```
cshields-oliveai, line 293: Given the current WIP implementation for "form like behavior" for whispers, the LDK will need a way to intercept components given as `UpdateWhisper` here in order to keep track of their value states.

ldk/javascript/.eslintrc.js
```diff
@@ -20,6 +20,8 @@ module.exports = {
     // https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-shadow.md
     'no-shadow': 'off',
     '@typescript-eslint/no-shadow': ['error'],
+    "no-use-before-define": "off",
+    "@typescript-eslint/no-use-before-define": ["error"]
```
richseviora, line 24: I'd much rather suppress the rule in a given instance unless we have a good reason to suppress it overall.

richseviora (changes requested):
Minor change to rule setting - otherwise looks straightforward!

ldk/javascript/.eslintrc.js
richseviora: Nope this was my error b/c I didn't read the change closely!

richseviora (approved):

PR #202, opened 2021-05-28
SIDE-2143: Form like component

cshields-oliveai:
This PR:
- adds a new LDK only `Form` component
- keeps track of values changing for embedded form components
- generates a submit button for each form (not currently configurable)
- sends component state back upon `onSubmit()` for form
- allows for multiple forms to be declared (no nesting of forms)

**notes**
- Need to make sure we are compensating for updating of whispers.
- Could potentially create a new card to extend form-like behavior (configuring submit button, validation, etc)


ldk/javascript/src/types/oliveHelps/index.d.ts
```diff
@@ -424,7 +433,13 @@ declare namespace OliveHelps {
     | Telephone
     | TextInput;
 
-  type Components = Box | ChildComponents | CollapseBox;
+  type Form = Component<WhisperComponentType.Form> & {
+    children: Array<ChildComponents>;
+    onSubmit: (values: Map<string, any>) => void;
+    // TODO: Make submit button configurable
```
RusGraf, line 439: Do you still need to do this todo?

ldk/javascript/src/types/oliveHelps/index.d.ts
```diff
@@ -248,6 +248,7 @@ declare namespace OliveHelps {
     Select = 'select',
     Telephone = 'telephone',
     TextInput = 'textInput',
+    Form = 'form',
```
RusGraf, line 251: I thought this d.ts file is representation of what we could communicate to/from goja runtime. And since Form is the LDK only concept, does it needs to be here?

ldk/javascript/src/whisper/form-parser.ts
```diff
@@ -0,0 +1,33 @@
...
+  };
+
+  whisper.components.forEach((component: Components) => {
+    if (isForm(component)) {
+      // Lift form components up
+      component.children.forEach(formChild => outgoingWhisper.components.push(formChild));
+
+      // Store form state
+      const ldkForm = new LdkForm(component.children);
+      ldkForms.push(ldkForm);
```
RusGraf, line 18: Do we need to delete form from state if whisper is closed? Wondering if this potentially could cause a leak or overlap with new form.

RusGraf: Hmm, perhaps not as your state is local to the parse function. Although should we be concerned about concurrency?

ldk/javascript/src/whisper/form.ts
```diff
@@ -0,0 +1,60 @@
...
+    private isOnChangeInput(child: any): boolean {
+        return (child.onChange !== undefined && child.name !== undefined);
+    }
+
+    private isOnSelectInput(child: any): boolean {
+        return (child.onSelect !== undefined && child.name !== undefined);
+    }
+
+    private setInitialComponentState(children: Array<any>) {
+        children.forEach((child: any) => {
```
RusGraf, line 30: do you need to check for children != undefined?

ldk/javascript/src/whisper/form.ts
```diff
@@ -0,0 +1,60 @@
...
+            if (this.isOnChangeInput(child)) {
+                this.componentState = this.componentState.set(child.name, child.value);
+            } else if (this.isOnSelectInput(child)) {
+                this.componentState = this.componentState.set(child.name, child.selected);
+            }
+        });
+    }
+
+    private registerListeners(children: Array<any>) {
+        children.forEach((child: any) => {
```
RusGraf, line 40: if (children)?

ldk/javascript/src/whisper/form.ts
```diff
@@ -0,0 +1,60 @@
...
+            if (this.isOnChangeInput(child)) {
+                this.componentState = this.componentState.set(child.name, child.value);
+            } else if (this.isOnSelectInput(child)) {
+                this.componentState = this.componentState.set(child.name, child.selected);
+            }
+        });
+    }
+
+    private registerListeners(children: Array<any>) {
```
RusGraf, line 39: Should it be Array<Components> instead of any?

ldk/javascript/src/types/oliveHelps/index.d.ts
cshields-oliveai: I'll remove it and create a card for it.

ldk/javascript/src/whisper/form.ts
```diff
@@ -0,0 +1,60 @@
...
+    getComponentState(): Map<string, any> {
+        return this.componentState;
+    }
+
+    private isOnChangeInput(child: any): boolean {
+        return (child.onChange !== undefined && child.name !== undefined);
+    }
+
+    private isOnSelectInput(child: any): boolean {
+        return (child.onSelect !== undefined && child.name !== undefined);
```
RusGraf, line 26: Will it be the same to have: (!child.onSelect && !child.name); ?

ldk/javascript/src/whisper/form-parser.ts
cshields-oliveai: Good thought. Let me see if I can isolate this issue with a test.

cshields-oliveai: Do you mean concurrency between forms? I don't believe so, each `Form` component will get its own internal state representation via one of these `LdkForm` object instances.

ldk/javascript/src/whisper/form.ts
```diff
@@ -0,0 +1,60 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { Components, Form, Whisper, WhisperComponentType } from '.';
+
+export class LdkForm {
+    children: Array<Components>;
```
cshields-oliveai, line 5: I do not need to store `children` here.

ldk/javascript/src/whisper/form.ts
cshields-oliveai: My function is `isOnChangeInput` so I believe I can change to `(child.onSelect && child.name)` where both need to evaluate as truthy.

ldk/javascript/src/whisper/form.ts
cshields-oliveai: I added a test to make sure of behavior with empty children collection.

ldk/javascript/src/whisper/form.ts
cshields-oliveai: This is not needed due to `.forEach`

ldk/javascript/src/whisper/form.ts
cshields-oliveai: I am not sure what you mean.

ldk/javascript/src/whisper/form.ts
cshields-oliveai: I will add another test for undefined.

ldk/javascript/src/whisper/form.ts
cshields-oliveai: Applied partway

ldk/javascript/src/types/oliveHelps/index.d.ts
richseviora: This is a good point and I think that including it here is incorrect because it implies that Goja can meaningfully do something with these fields and types, and it cannot. Best case it will ignore. Worst case it will error out.

ldk/javascript/src/types/oliveHelps/index.d.ts
```diff
@@ -313,13 +314,15 @@ declare namespace OliveHelps {
     tooltip?: string;
     value: boolean;
     onChange: WhisperHandlerWithParam<boolean>;
+    name?: string;
```
richseviora, line 317: Also drop the `name` properties from this file.

ldk/javascript/src/whisper/form.ts
richseviora: probably could express this as:
```suggestion
        children?.forEach((child: any) => {
```

ldk/javascript/src/whisper/form.ts
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    private isOnSelectInput(child: any): boolean {
+        return (child.onSelect && child.name);
+    }
+
+    private setInitialComponentState(children: Array<any>) {
+        children.forEach((child: any) => {
+            if (this.isOnChangeInput(child)) {
+                this.componentState = this.componentState.set(child.name, child.value);
```
richseviora, line 32: The `Map` object here is being mutated when `Map.set` is called. Shouldn't need to set the `componentState` property again.

ldk/javascript/src/whisper/form.ts
```diff
@@ -0,0 +1,60 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
richseviora, line 1: Where we're disabling a rule we should explain why we're doing it. 

Here I think we can instead set the Map's value type to `string|number|boolean`, so the signature becomes then `Map<string,string|number|boolean>`.

richseviora (changes requested):

ldk/javascript/src/types/oliveHelps/index.d.ts
cshields-oliveai: I do not disagree. I will have to create some kind of mapper to go from our LDK `NewWhisper` type to the `NewWhisper` `d.ts` contract with OH.

ldk/javascript/src/whisper/form.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    constructor(children: Array<Components>) {
+        this.componentState = new Map<string, any>();
+        this.children = children;
+
+        this.setInitialComponentState(this.children);
+        this.registerListeners(this.children);
+    }
+
+    getComponentState(): Map<string, any> {
```
cshields-oliveai, line 17: I was testing this further, and I noticed that there is a race condition when retrieving the current component state upon form submission. We are not guaranteed to get `onChange()` called before we retrieve the current state for each component collection.

ldk/javascript/src/whisper/form.ts
cshields-oliveai: Removed double assignment.

ldk/javascript/src/whisper/form.ts
cshields-oliveai: Good idea. Updated.

ldk/javascript/src/whisper/whisper.test.ts
```diff
@@ -9,7 +8,7 @@ describe('Whisper', () => {
   });
 
   describe('create', () => {
-    it('Creates a whisper', () => {
+    it('creates a whisper', async () => {
```
cshields-oliveai, line 11: `async`

ldk/javascript/src/whisper/index.ts
```diff
@@ -264,6 +284,33 @@ export interface WhisperAptitude {
   create(whisper: NewWhisper): Promise<Whisper>;
 }
 
-export function create(whisper: NewWhisper): Promise<Whisper> {
-  return promisifyWithParam(whisper, oliveHelps.whisper.create);
+function addUpdateHandler(whisper: OliveHelps.Whisper): Whisper {
```
RusGraf, line 287: Will it be cleaner to have a whole 'addUpdateHandler' function to be moved to the mapper? Basically you doing a 'mapToWhisper(whisper: OliveHelps.Whisper)'

ldk/javascript/src/whisper/index.ts
```diff
@@ -264,6 +284,33 @@ export interface WhisperAptitude {
...
+  externalWhisper.update = (updateWhisper: UpdateWhisper, cb?: (err: Error) => void) => {
+    updateFunc(mapToInternalUpdateWhisper(updateWhisper), cb);
+  };
+  return externalWhisper;
 }
+
+export function create(whisper: NewWhisper): Promise<Whisper> {
+  const internalWhisper: OliveHelps.NewWhisper = mapToInternalWhisper(whisper);
+
+  return new Promise((resolve, reject) => {
```
RusGraf, line 299: I think after moving addUpdateHandler to mapper you should be able to use: 'promisifyMappedWithParam'.

ldk/javascript/src/whisper/whisper-mapper.ts
```diff
@@ -0,0 +1,34 @@
+import { NewWhisper, UpdateWhisper } from ".";
+import { whisper } from "..";
+import { parseForm } from "./form-parser";
+
+/* eslint-disable-next-line @typescript-eslint/no-explicit-any */ // We have to coerce any type here to force conversion of whisper.NewWhisper to OliveHelps.NewWhisper 
+export function convertComponentType(whisperComponentType: whisper.WhisperComponentType): any {
```
RusGraf, line 6: Not sure if this is really a mapper or conversion function. Looks more like a validation. It is strange that we will throw an error here. Perhaps we could introduce a validation layer?

RusGraf (approved):

PR #210, opened 2021-06-03
[SIDE-2274] Replace p with div to address facebook/react#5479

richseviora:
Disappearing text issue caused by nested `p` tag reconciliation issue. Also making this branch trigger doc updates to test.


cshields-oliveai (approved):

.github/workflows/prettier.yml
```diff
@@ -0,0 +1,18 @@
+name: Prettier
```
flowermm, line 1: Happy to see Prettier added!

ldk/javascript/src/webpack/config.ts
```diff
@@ -2,9 +2,9 @@ import * as webpack from 'webpack';
 import * as path from 'path';
 import Terser from 'terser-webpack-plugin';
 import { generateBanner } from './generate-banner';
-import { LdkSettings } from './ldk-settings';
+import { LdkSettings } from './ldk-settings'; // Need to dynamically refer to Loop's package.json
 
-/* eslint-disable-next-line */ // Need to dynamically refer to Loop's package.json
```
flowermm, line 7: I realize that the bulk of this PR is prettier doing its thing. Is it worth keeping why we need to disable the linter? Maybe change the eslint disable to reference the exact rule we are breaking

flowermm (approved):
Glad to see the Gatsby/React issue was a simple fix! Also excited for prettier to help keep things looking uniform 👍

.github/workflows/prettier.yml
richseviora: In addition to being a Typescript fan boy I am _also_ a Prettier fanboy 😀

ldk/javascript/src/webpack/config.ts
richseviora: Ooo yes, great point, I try to document the specific rule and why I'm suppressing it, and I should do that here too.

richseviora: Is done!

PR #212, opened 2021-06-07
SIDE-1978: Make ListPair Label text copyable

epiccoleman:
Adds an optional configuration value on the ListPair Whisper Component to enable copying of Label text.


DanielWagersOlive (approved):

ldk/javascript/examples/self-test-loop/src/tests/index.ts
```diff
@@ -881,6 +882,60 @@ export const linkWhisper = (): Promise<boolean> =>
     });
   });
 
+
+export const listPairWhisperCopyable = (): Promise<boolean> =>
```
cshields-oliveai, line 886: If you wanted to get ambitious, you could use the clipboard aptitude to make sure that the text you had in the list pair was copied.

cshields-oliveai (approved):

ldk/javascript/examples/self-test-loop/src/tests/index.ts
```diff
@@ -13,6 +13,7 @@ import {
 } from '@oliveai/ldk';
 
 import { Cancellable } from '@oliveai/ldk/dist/cancellable';
+import { Urgency } from '../../../../dist/whisper';
```
RusGraf, line 16: will it be better to import this from @oliveai/ldk/dist/whisper to make sure it is properly exported?

ldk/javascript/examples/self-test-loop/src/tests/index.ts
epiccoleman: I like that idea, I'll give that a shot.

ldk/javascript/examples/self-test-loop/src/tests/index.ts
epiccoleman: Probably! I think the editor helped me out there. I'll update that.

ldk/javascript/src/types/oliveHelps/index.d.ts
```diff
@@ -290,7 +290,7 @@ declare namespace OliveHelps {
   interface Whisper {
     id: string;
     close: Readable<undefined>;
-    update(whisper: UpdateWhisper, cb?: (err: Error) => void): void 
+    update(whisper: UpdateWhisper, cb?: (err: Error) => void): void
```
RusGraf, line 293: Are you missing semicolon?

RusGraf (approved):

ldk/javascript/src/types/oliveHelps/index.d.ts
epiccoleman: I think this is where VSCode knocked off a trailing space. I'll check to be sure though.

ldk/javascript/examples/self-test-loop/src/tests/index.ts
epiccoleman: Did it. Good idea.

PR #214, opened 2021-06-08
[SIDE-2331] Rename alignment to justifyContent

richseviora:
This PR:

- Renames the `alignment` property on boxes to `justifyContent` to be consistent with CSS naming.
- Converts enums in the declaration file to types b/c the enums don't actually exist.
- More type refactoring.


ldk/javascript/examples/self-test-loop/src/tests/index.ts
```diff
@@ -251,14 +251,14 @@ export const testMarkdownWhisper = (): Promise<boolean> =>
...
-            alignment: whisper.Alignment.SpaceEvenly,
+            justifyContent: whisper.JustifyContent.SpaceEvenly,
             direction: whisper.Direction.Horizontal,
             children: [
               {
                 label: `No`,
                 onClick: () => {
                   form.close((error) => console.error(error));
-                  reject(true);
+                  reject(Error);
```
RusGraf, line 261: I'm curious to understand what would happen if you call reject(Error); Should it be reject(new Error('Some message'));?

ldk/javascript/src/promisify.ts
```diff
@@ -8,14 +8,15 @@ function handleCaughtError(reject: (reason?: any) => void, error: Error) {
   reject(error);
 }
 
-export function promisifyMappedWithParam<TParam, TInternalOut, TExternalOut>(
-  param: TParam,
+export function promisifyMappedBothWithParams<TParamIn, TParamOut, TInternalOut, TExternalOut>(
+  param: TParamIn,
+  paramMap: Mapper<TParamIn, TParamOut>,
```
RusGraf, line 13: Will it be simpler to call mapper and then call promisifyMappedWithParam() instead of expanding this layer?

ldk/javascript/src/types/oliveHelps/index.d.ts
```diff
@@ -300,56 +283,56 @@ declare namespace OliveHelps {
   type WhisperHandler = (error: Error | undefined, whisper: Whisper) => void;
   type WhisperHandlerWithParam<T> = (error: Error | undefined, param: T, whisper: Whisper) => void;
 
-  type Button = Component<WhisperComponentType.Button> & {
+  type Button = Component<'button'> & {
```
RusGraf, line 286: Why can it not be "WhisperComponentType.button"? I thought having string defined in one place and using strongly typed param would reduce potential future issues.

ldk/javascript/src/whisper/index.ts
```diff
@@ -230,19 +238,30 @@ export type ChildComponents =
...
-export type Components = Box | ChildComponents | CollapseBox;
+export type Box = WhisperComponent<WhisperComponentType.Box> & {
+  justifyContent: JustifyContent;
+  children: Array<ChildComponents>;
+  direction: Direction;
+};
+
+export type Components = DeprecatedBox | Box | ChildComponents | CollapseBox;
+
+export type BoxChildComponents = ChildComponents | Box | DeprecatedBox;
```
RusGraf, line 264: I'm a bit puzzled here. Box and DeprecatedBox already defined in the ChildComponents enum. Then why do we need to add them here?

ldk/javascript/src/whisper/index.ts
```diff
@@ -230,19 +238,30 @@ export type ChildComponents =
   | Telephone
   | TextInput;
 
-export declare type CollapseBox = WhisperComponent<WhisperComponentType.CollapseBox> & {
+export type CollapseBox = WhisperComponent<WhisperComponentType.CollapseBox> & {
   children: Array<ChildComponents>;
```
RusGraf, line 242: If you add Box into the ChildComponents that will allow loop developer to add a Box inside of the CollapseBox. I thought requirement to allow Box in the Box was only for the Box component and not the CollapseBox.

ldk/javascript/src/whisper/index.ts
```diff
@@ -265,6 +284,146 @@ export interface WhisperAptitude {
   create(whisper: NewWhisper): Promise<Whisper>;
 }
 
+function convertChildComponent(component: BoxChildComponents): OliveHelps.ChildComponents {
```
RusGraf, line 287: Will it be cleaner to move it to a mapper layer?

ldk/javascript/src/promisify.ts
cshields-oliveai: I still vote that this is not a sustainable pattern..

ldk/javascript/src/types/oliveHelps/index.d.ts
cshields-oliveai: The Loop author continues to use `whisper.WhisperComponentType.Button` I do believe.

ldk/javascript/src/whisper/index.ts
cshields-oliveai: My vote is a new file: `src/whisper/whisper-mapper.ts`. I envision a future where most aptitudes could end up with an associated mapper.

ldk/javascript/src/whisper/index.ts
```diff
@@ -265,6 +284,146 @@ export interface WhisperAptitude {
...
+        open: component.open,
+        children: component.children.map(convertChildComponent),
+        type: WhisperComponentType.CollapseBox,
+      };
+    default:
+      return convertChildComponent(component);
+  }
+}
+
+function convertToGojaWhisper(whisper: UpdateWhisper): OliveHelps.UpdateWhisper;
```
cshields-oliveai, line 395: I like the naming convention `internal` which would signify `OliveHelps` flavor of types and `external` that would signify our LDK aptitude types.

ldk/javascript/src/whisper/index.ts
```diff
@@ -265,6 +284,146 @@ export interface WhisperAptitude {
...
+        components: whisper.components.map(convertComponent),
+      };
+}
+
+function convertGojaWhisper(whisper: OliveHelps.Whisper): Whisper {
+  return {
+    id: whisper.id,
+    close: whisper.close,
+    update(updateWhisper: UpdateWhisper, cb): void {
+      whisper.update(convertToGojaWhisper(updateWhisper), cb);
```
cshields-oliveai, line 417: It's very important that `whisper.update()` is called on a converted whispers. Can a test be added for this?

ldk/javascript/examples/self-test-loop/src/tests/index.ts
richseviora: You're correct! Doing this would reject the promise with the Error class, which doesn't make a lick of sense. I'm not sure why I changed that - will fix :)

ldk/javascript/src/types/oliveHelps/index.d.ts
richseviora: OK so this is just the change on the internal representation, so it doesn't affect the Loop author's experience like Cory mentioned.

It can't be `WhisperComponentType.Button` because that enum doesn't exist (never did). What I could've done was something like this:
```typescript
type WhisperTypePropMarkdown = 'markdown';
type WhisperTypePropButton = 'button';
...
type WhisperComponentType = 
  | WhisperTypePropMarkdown
  | WhisperTypePropButton;
```

But when I started playing around with it it felt pretty awkward to use and type, so I ended up defining `OliveHelps.WhisperComponentType` as a union type of all the possible strings:
```typescript
type WhisperComponentType = 
  | 'markdown'
  | 'button';
```

So while the param is still strongly typed, the difference is that a) it doesn't refer to something that doesn't exist, and b) the only way this becomes a problem is if we repurpose one of the existing strings for something else entirely, which we won't do.

ldk/javascript/src/whisper/index.ts
richseviora: Oops! Good catch!

ldk/javascript/src/whisper/index.ts
richseviora: Ahhh - gotcha, OK, let me fix that

ldk/javascript/src/promisify.ts
richseviora: @RusGraf I'm not sure what you mean by calling mapper here; can you clarify?

ldk/javascript/src/whisper/index.ts
richseviora: Depends on what you mean by a mapper layer - we can move the conversion logic into a separate file, absolutely.

ldk/javascript/src/whisper/index.ts
richseviora: Sure! I'll make that change

ldk/javascript/src/whisper/index.ts
richseviora: Sure!

ldk/javascript/src/whisper/index.ts
richseviora: I've moved the conversion logic into `convert.ts`. I've also moved the types into `types.ts` to avoid any cyclic dependency issues down the road.

ldk/javascript/src/whisper/index.ts
richseviora: Added!

ldk/javascript/src/whisper/index.ts
richseviora: Fixed!

ldk/javascript/src/whisper/index.ts
richseviora: Fixed!

ldk/javascript/src/promisify.ts
RusGraf: Yes, what I mean is the reason why we calling mapper as part of the prdmosify in other places is that we are mapping return parameters and we are trying to resolve with the mapped value. I'm wondering if it will be better if we do not grow generic promisify layer by adding extra complexity. Since we are trying to map input parameter could we just do mapping upfront before we call promisify?

Something like this:
`export function create(whisper: NewWhisper): Promise<Whisper> {
  const internalWhisper = convertToInternalWhisper(whisper);
  return promisifyMappedWithParam(
    whisper,
    convertToExternalWhisper,
    oliveHelps.whisper.create,
  );
}`

ldk/javascript/src/types/oliveHelps/index.d.ts
RusGraf: I understand your point. I think what took my attention was that we are using same strings in 2 different places like ('button')
`type Button = Component<'button'> & {` 
and 
`type WhisperComponentType =
    | 'box'
    | 'button'`

Wondering if it could lead to an issue if we decide to update a string to something else. Preferably it would be defined in 1 place and we could have them strongly typed. But this is minor and I don't think I'm strongly stand on it ,so I will leave it up to you. Just want to make sure I explain better what I meant.

ldk/javascript/src/whisper/index.ts
RusGraf: Yes, that is what I meant ;) Also, for consistency, could we call it map instead of convert? We already have mapper.ts and I think Cory is adding whisper-mapper.ts or something along this line. I think we should stick with 1 naming convention: convert or map.

ldk/javascript/src/promisify.ts
richseviora: I don't think this is adding any material complexity, though this does add a small abstraction. If I didn't feel we'd be likely to be converting more values I wouldn't go for this either. But I think we are likely to be converting more incoming values and here's why I think this logic makes sense here:

Putting all of our whispers through a standard handling layer allows us to standardize the LDK <-> OH interactions and makes the code IMO more maintainable in the long term. 

What I'd like to prevent is a scenario where a few years down the line we have 100 different capabilities that are doing some mapping conversions.  We identify a need to time the mapping process for whatever reason. With a common function like this we can easily instrument that logic. Without a common function like this, we'd have to instrument each of the functions individually.

Another potential scenario is that we decide at some point that we want all errors that rise during the mapping of incoming parameters to reject a promise instead of throwing an error (or vice versa). With a common process abstraction we can change the process easily.

ldk/javascript/src/types/oliveHelps/index.d.ts
richseviora: I think I basically agree with everything you said above in analyzing the problem and potential pitfalls. It could go either way, and it's a judgment call and if the team feels strongly about the other way I'll go with that too. I think my logic here is that these string values are very unlikely to be mutated. We will never rename a `'button'` to `'specialButton'`, as an example. Adding a separate type declaration adds a bit of indirection, so you need to unpeel that to get the underlying value.

ldk/javascript/src/promisify.ts
RusGraf: Yeah, I'm in support of the intent. The angle I was looking at is that promisify is not following SRP which is making it less maintainable. But perhaps it is just my view on it and I don't think we have any functional issues here so we could move on.

RusGraf (approved):

richseviora:
OK I've reviewed the content and confirmed that this will be a minor change only.

cshields-oliveai (approved):

epiccoleman (approved):

PR #219, opened 2021-06-10
SIDE-2479: Update User Aptitude Documentation

epiccoleman:
This PR updates the User Aptitude documentation with the some additional information, most significantly the list of JWT claims and the public key which can be used to validate the signatures of User Aptitude JWTs.


EddieAtOlive:
This looks like a good improvement over the current docs for sure. Thanks!

One thing that might be worth considering is what path we use for the JWT's public key. Looks like it's currently set to `oliveai.dev/static/user-aptitude-jwt-rsa.pub` and we will eventually expose this through the `auth.oliveai.com` domain. It might be worth making the path on oliveai.dev mimic the path we'll eventually use. So for example, if we plan to make it: `auth.oliveai.com/public-keys/olive-helps/user-aptitude/jwt-rsa.pub`, then maybe we'd want the current one to be: `oliveai.dev/static/public-keys/olive-helps/user-aptitude/jwt-rsa.pub`. It's not a big deal, but it would be nice to settle on a consistent path now if we can do that quickly.

DanielWagersOlive (approved):

docs/src/markdown-pages/aptitudes/user.createJWT.md
```diff
@@ -2,5 +2,20 @@
...
 links_js: "jwt"
 ---
-Receive a [JWT](https://jwt.io) with the currently logged in Olive Helps user's email address in the `subject` claim, the Loop ID in the `audience` claim, and a 5-minute `expiration`.
+Receive a [JWT](https://jwt.io) identifying the currently logged in Olive Helps user.
 
+## JWT Claims
+The claims in the JWT are as follows:
+
+Standard Claims:
+`sub` (Subject): contains the unique User ID of the current Olive Helps user
```
BrettMugglinOlive, line 11: Super nitpick, but should these sentences here have periods at the end of them?

BrettMugglinOlive (approved):

EddieAtOlive:
it's fine if this is intentional, but I wanted to double check that you do in fact want user-aptitude repeated twice here:
`user-aptitude/user-aptitude-jwt-rsa.pub`. Were you going for `user-aptitude/jwt-rsa.pub` or did you want that repeated?

docs/src/markdown-pages/aptitudes/user.createJWT.md
```diff
@@ -2,5 +2,20 @@
...
 links_js: "jwt"
 ---
-Receive a [JWT](https://jwt.io) with the currently logged in Olive Helps user's email address in the `subject` claim, the Loop ID in the `audience` claim, and a 5-minute `expiration`.
+Receive a [JWT](https://jwt.io) identifying the currently logged in Olive Helps user.
 
+## JWT Claims
+The claims in the JWT are as follows:
+
+Standard Claims:
+`sub` (Subject): contains the unique User ID of the current Olive Helps user.
```
richseviora, line 11: Markdown treats single new lines as standard space so they'll run together. To present them in a list like this you'll probably want to format them as markdown lists, or maybe as a table.

richseviora (changes requested):
Something you probably didn't catch due to the Gatsby crashing issue :)

docs/src/markdown-pages/aptitudes/user.createJWT.md
epiccoleman: Fixed.

richseviora (approved):

PR #220, opened 2021-06-10
Side 2173 make all uint8array take string

RusGraf:


MaggieMaSS (approved):

ldk/javascript/package.json
```diff
@@ -18,6 +18,7 @@
   ],
   "scripts": {
     "build": "rm -rf dist && tsc",
+    "prepush": "npm run build && npm run test && npm run lint",
```
cshields-oliveai, line 21: *high five*

RusGraf: hehe, 1 step at a time.

ldk/javascript/src/utils/network.mapper.ts
```diff
@@ -0,0 +1,99 @@
+import { TextDecoder } from 'text-encoding-shim';
```
cshields-oliveai, line 1: Would it be worthwhile to move this to `src/network/network.mapper.ts`? It's becoming apparent that we will need mappers for many aptitudes, and I thought putting individual mappers closer to them might make sense.

cshields-oliveai (approved):

ldk/javascript/src/utils/network.mapper.ts
RusGraf: I'm actually not sure what would be the most idiomatic way here. I saw applications with both structures. Do you know specifically why would you do one way vs another?

ldk/javascript/package.json
cshields-oliveai: A thought also occurred to me - what about including building the `examples` dir as well to help prevent those projects from going stale?

RusGraf: Hmm, I think it is a good idea to run them, but I'm wondering where would be best to do it. Perhaps it is better fits to do it in github actions via hooks?

cshields-oliveai: Maybe we can create a new user story for this.

PR #227, opened 2021-06-15
Side 2425 component tooltip

RusGraf:
Creating tooltips for button, markdown and message controls.


ldk/javascript/examples/self-test-loop/src/selfTest.ts
```diff
@@ -44,6 +35,18 @@ import {
   fileExists,
 } from './tests';
 
+import {
+  testBoxInTheBox,
```
cshields-oliveai, line 39: :D

ldk/javascript/examples/self-test-loop/src/tests/whisper.ts
```diff
@@ -0,0 +1,700 @@
...
+                  form.close((error) => console.log(error));
+                  reject(false);
+                },
+              },
+            ],
+          },
+        ],
+      });
+    } catch (error) {
+      console.error(error);
```
cshields-oliveai, line 307: Might want to close the whispers as well upon rejection.

ldk/javascript/examples/self-test-loop/src/tests/whisper.ts
```diff
@@ -0,0 +1,700 @@
+import { clipboard, whisper } from '@oliveai/ldk';
```
cshields-oliveai, line 1: Is this diff not showing these tests moved from `self-test-loop/src/tests/index.ts` or were they not removed?

cshields-oliveai (approved):

ldk/javascript/examples/self-test-loop/src/tests/whisper.ts
cshields-oliveai: Even though I don't believe you added this.

RusGraf: I added a timeout form close to close the whisper if no action from the user received. Although we cant do it in the catch as it means that whisper is not properly created and there is no instance to close.

ldk/javascript/examples/self-test-loop/src/tests/whisper.ts
RusGraf: Well, looks like merging from main brought them back. Let me fix it.

RusGraf: Cleaned up

epiccoleman (approved):

PR #229, opened 2021-06-15
SIDE-2512: Updated multiline markdown example, and updated docs

cshields-oliveai:


docs/src/markdown-pages/aptitudes/whisper.create.md
```diff
@@ -16,6 +16,12 @@ The following component types are available:
 * ListPair - This component shows a two column view of information typically used for lists of information.
 * Markdown - Renders a message as Markdown.
     * Markdown syntax for the Markdown Component Type is defined by to the [CommonMark](https://commonmark.org/) specification. 
+    * Tip: if using a Template Literal for specifying markdown `body`, it should be noted that whitespace will be preserved when rendering markdown. As a consequence of this, markdown specified with tab whitespace (spaces, tabs, etc.) at the beginning of a line will **not** be interpreted as markdown. One way to avoid this situation, is to use a regular expression replacement function with Template Literals like `.replace(/(\n)\s+/g, '$1')` to replace all line breaks followed by whitespace with only a line break. For example:
```
richseviora, line 19: > will **not** be interpreted as markdown

I thought the problem was that leading spaces (including those suggested by code indentation) were being treated as spaces by markdown. And depending on the amount of spacing, that could trigger some weird behaviour.

I think the replace approach here is good in a jam, but it could cause some problems. Consider:

```js
  const str = `
  # I am a header.
  I am a line.
  * I am a list item.
  * I am a list item.
    * I am an indented list item.
`;
```

The replace call would remove the indentation from the indented list item and flatten it out. 

Instead I'd suggest using something like the `stripIndent` tagged template literal function in the [common-tags](https://www.npmjs.com/package/common-tags#stripindent) library.

richseviora (changes requested):

ldk/javascript/examples/self-test-loop/src/tests/index.ts
```diff
@@ -197,7 +197,7 @@ export const testMarkdownWhisper = (): Promise<boolean> =>
               | - | - |
               | Row 1 Col 1 | Row 1 Col 2 |
               | Row 2 Col 1 | Row 2 Col 2 |
```
RusGraf, line 199: Should we remove tab space before each line in the markdown? Perhaps I didn't fully understood the issue.

richseviora (approved):

RusGraf (approved):

PR #232, opened 2021-06-17
SIDE-2450: Add examples to Websocket docs

epiccoleman:


docs/src/markdown-pages/aptitudes/network.websocket.md
```diff
@@ -2,4 +2,33 @@
...
+    // do something with message contents
+})
+```
+
+Write messages to the WebSocket with `writeMessage`:
+```
+await socket.writeMessage("Hello WebSocket!")
+```
+
+Set a callback for when the WebSocket is closed:
```
flowermm, line 29: Is the user responsible for closing their socket? I'm assuming there is a socket.close()?

flowermm (approved):

docs/src/markdown-pages/aptitudes/network.websocket.md
RusGraf: That is correct, you could do socket.close(). Although your socket could also be closed by the server and we want to give user a way to subscribe if this event will occur.

RusGraf: @epiccoleman I think Michelle brought up a good point. Should we indicate that user could call socket.close() to kill the socket connection?

RusGraf (approved):

PR #236, opened 2021-06-21
Side 2568 update example whisper imports

RusGraf:


ldk/javascript/examples/javascript-loop/index.js
```diff
@@ -1,4 +1,5 @@
 import { clipboard, whisper } from '@oliveai/ldk';
+import { WhisperComponentType } from '@oliveai/ldk/dist/whisper/types';
```
MaggieMaSS, line 2: Is it gonna be better if `import { Markdown, Button, ... } from '@oliveai/ldk/dist/whisper/types/WhisperComponentType';`?  Then we can directly use `type: Markdown` inside our component

DanielWagersOlive (approved):

epiccoleman (approved):

PR #243, opened 2021-06-23
Add CollapseBox.onClick to the mapper function

BrettMugglinOlive:
Fixes an issue with building the self test loop


DanielWagersOlive (approved):

ldk/javascript/src/whisper/mapper.ts
```diff
@@ -120,6 +120,19 @@ export function mapToInternalChildComponent(
 export function mapToInternalComponent(component: Component): OliveHelps.Components {
   switch (component.type) {
     case WhisperComponentType.CollapseBox:
+      // eslint-disable-next-line
```
flowermm, line 123: Can you use the specific eslint rule disable for posterity :)

flowermm (approved):

ldk/javascript/src/whisper/mapper.ts
BrettMugglinOlive: Updated

PR #244, opened 2021-06-23
Run Webpack against node_modules

epiccoleman:
This fixes an issue introduced in 3.0.1 where the Webpack config used for Loop compilation was not appropriately converting ES6 code from the LDK to ES5 for use by Helps.


BrettMugglinOlive:
Does this still need to be merged? Or have we addressed this in a different PR?

flowermm:
Brett, this is the same file that has been approved in https://github.com/open-olive/loop-development-kit/pull/246/files

epiccoleman:
@BrettMugglinOlive this was addressed in #246 , which also includes the version bump. I'll close this PR.

PR #245, opened 2021-06-23
SIDE: 2143 component state

cshields-oliveai:
This PR:
- keeps track of values changing for whisper embedded form components
- sends component state back as `componentState` on whisper
- allows for multi-whisper component state

### DIFF
--- a/ldk/javascript/examples/self-test-loop/src/selfTest.ts
+++ b/ldk/javascript/examples/self-test-loop/src/selfTest.ts
@@ -52,7 +52,12 @@ import {
   tooltips,
   onBlurTest,
 } from './tests/whisper';
-import { basicWhisperUpdate, updateCollapseState, updateOnChange } from './tests/whisper-update';
+import {
+  basicWhisperUpdate,
+  updateCollapseState,
+  updateOnChange,
+  whisperStateOnChange,
+} from './tests/whisper-update';
 
 const testConfig: { [key: string]: TestGroup } = {
   clipboard: new TestGroup('Clipboard Aptitude', [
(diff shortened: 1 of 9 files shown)


ldk/javascript/examples/self-test-loop/src/selfTest.ts
```diff
@@ -275,6 +280,12 @@ const testConfig: { [key: string]: TestGroup } = {
       20000,
       `Did the whisper update correctly?`,
     ),
+    new LoopTest(
+      'Whisper Update - Automated OnChange',
+      whisperStateOnChange,
+      20000,
```
RusGraf, line 286: If no action needed, do we need to have it 20sec? in case of failure it feels like a long timeout.

ldk/javascript/examples/self-test-loop/src/tests/whisper-update.ts
```diff
@@ -1,56 +1,29 @@
...
   ChildComponents,
-  Components,
-  Markdown,
+  Component,
+  NumberInput,
+  Select,
   TextInput,
   Whisper,
+  WhisperComponentType,
 } from '@oliveai/ldk/dist/whisper';
```
RusGraf, line 11: make it `@oliveai/ldk/dist/whisper/types` please. We currently reexporting the types, but I hope we will stop at some point and it is better to reduce the future change here.

ldk/javascript/examples/self-test-loop/src/tests/whisper-update.ts
```diff
@@ -1,56 +1,29 @@
...
-): Array<Components> => [
+): Array<Component> => [
   {
     type: whisper.WhisperComponentType.Message,
     body: prompt,
   },
   {
     type: whisper.WhisperComponentType.Box,
-    alignment: whisper.Alignment.SpaceAround,
+    alignment: whisper.JustifyContent.SpaceAround,
```
RusGraf, line 26: Could you please get JustifyContent from types and not whisper?

RusGraf:
I left a couple of minor things, you could fix them or not per your discretion. We also talked about adding documentation into docs. Looks good otherwise.

RusGraf (approved):

epiccoleman (approved):

PR #261, opened 2021-07-01
HELPS-756 Add to scripts to publish to Verdaccio

BrettMugglinOlive:


ldk/javascript/examples/self-test-loop/package.json
```diff
@@ -58,7 +58,7 @@
   "dependencies": {
     "@babel/runtime": "7.13.10",
     "@babel/runtime-corejs2": "7.13.10",
-    "@oliveai/ldk": "file:../..",
```
RusGraf, line 61: if we pointing to a specific version, should we move selfTestLoop out of the examples folder? I think after this change there is no benefit for it to be there.

BrettMugglinOlive: Possibly. I'm not quite sure where we should move it though...maybe to a different github repo or something?

ldk/javascript/examples/self-test-loop/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.ts",
   "scripts": {
     "build": "webpack",
-    "build-ldk": "cd ../.. && npm i && rm -rf dist && tsc",
+    "build-ldk": "cd ../.. && npm i && rm -rf dist && tsc && && npm unpublish @oliveai/ldk -f --registry http://localhost:4873 && npm publish --registry http://localhost:4873",
```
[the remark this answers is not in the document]
BrettMugglinOlive: The script here is supposed to be used with the Verdaccio tool I had talked about, which the way to get that set up was described [here](https://docs.google.com/document/d/1kTmNwmWOJcAU1sjNIcCpDbJHfhKkmrjkt9ZuOUqu3_M/edit)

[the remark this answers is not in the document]
RusGraf: So each time we build selfTestLoop locally we will be publishing to Verdaccio. I'm not fully understand why do we need to do it on build? Perhaps I missed some conversation and just confused here, but I would think we could have scripts like:
`
"publish": "npm unpublish @oliveai/ldk -f --registry http://localhost:4873 && npm publish --registry http://localhost:4873",
"install": "build-ldk && publish"
`

ldk/javascript/examples/self-test-loop/package.json
RusGraf: Perhaps we could chat about it to make sure we all aligned on if this could cause issues in the future. I thought selfTestLoop ideally will become an automated testing framework which will be running in CI. In this case it will need to be testing compiled version of LDK and not a specific released version. Developing locally we could override it but it wont work for CI. Perhaps publishing could point to a specific version before building the selfTestLoop, not sure. Also, I may jumping too far ahead on this topic, but want to bring it up to make sure we are aligning with a future vision.

ldk/javascript/examples/self-test-loop/package.json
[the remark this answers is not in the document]
BrettMugglinOlive: You shouldn't need to republish to Verdaccio unless there was a change in the Javascript LDK itself. Changes that were made to just self test, like modifying a test or adding one, still just need `npm run build`, which doesn't republish.

ldk/javascript/examples/self-test-loop/package.json
BrettMugglinOlive: I guess I thought the point of this was to specifically make the self test loop reference a compiled version of the LDK, rather than a relative path. Cause the compiled version had a bug that we didn't catch cause we were using the simlink.

RusGraf (approved):

flowermm (approved):

PR #271, opened 2021-07-12
HELPS-121 Add analytics to developer hub

BrettMugglinOlive:


epiccoleman (approved):

flowermm (changes requested):
Let's add tracking to "get started", "access the LDK", and "LDK" menu item as well. And maybe the additional resources? Helps us understand if people are viewing documentation or not via this launch page. I realize this is basically asking to track every button - there isn't a good way to understand how people are getting to the docs currently. This is the one time I wish we had Pendo flow (though maybe GA has an equivalent) - to understand if people are clicking more buttons if they don't find what they want on the first time 🤔

docs/src/components/menu/v2menu.tsx
```diff
@@ -9,7 +9,16 @@ interface LinksProps {
...
 const Links: React.FunctionComponent<LinksProps> = (props) => (
   <ul className={props.listClass}>
-    <li className={props.itemClass}>
+    <li
+      className={props.itemClass}
+      onClick={() => {
+        window.gtag('event', 'Link out', {
+          event_category: 'Navigation',
+          event_label: 'Github Pages',
+          value: 0,
```
flowermm, line 18: any way to get info on what github page it is for value?

BrettMugglinOlive: Well it leads to the root page of of our autogenerated Github pages docs [here](https://open-olive.github.io/loop-development-kit/ldk/javascript/). So I thought this was appropriate, but I can change it to something else

flowermm: works for me! ha

flowermm (approved):

PR #275, opened 2021-07-14
[HELPS-853] Add Dropzone Component

richseviora:
This PR adds the dropzone component.

### DIFF
--- a/ldk/javascript/src/whisper/mapper.ts
+++ b/ldk/javascript/src/whisper/mapper.ts
@@ -112,6 +112,13 @@ export function mapToInternalChildComponent(
     case WhisperComponentType.Markdown:
     case WhisperComponentType.Message:
       return component;
+    case WhisperComponentType.DropZone:
+      return {
+        ...component,
+        onDrop: (error, param, whisper) => {
+          component.onDrop(error, param, mapToExternalWhisper(whisper, stateMap))
+        }
+      };
     case WhisperComponentType.Number:
       if (component.id && component.value) {
         stateMap.set(component.id, component.value);
@@ -207,7 +214,7 @@ export function mapToInternalChildComponent(
         },
       } as OliveHelps.DateTimeInput;
     default:
-      throw new Error('Unexpected component type');
+      throw new Error(`Unexpected component type: ${(component as any)?.type}`);
   }
 }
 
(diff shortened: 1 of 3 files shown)


ldk/javascript/src/whisper/mapper.ts
```diff
@@ -112,6 +112,13 @@ export function mapToInternalChildComponent(
     case WhisperComponentType.Markdown:
     case WhisperComponentType.Message:
       return component;
+    case WhisperComponentType.DropZone:
+      return {
+        ...component,
+        onDrop: (error, param, whisper) => {
```
RusGraf, line 118: Do we need to set a 'path' value in the state map so it will be returned back to loop author as part of any onSelect/onChange/onDrop events?

RusGraf:
How will we test this component? Will it make sense to add selfTestLoop test to validate validation, tooltip and general functionality?

ldk/javascript/src/whisper/mapper.ts
richseviora: Maybe! TBH I'm not sure how we'll handle the state of this specific component. Unlike our other state values, these file objects are representations of something the user can kinda interact with, and not the actual value. But this will be handled in HELPS-854 and HELPS-856.

richseviora:
Testing isn't in scope for this ticket; this is jsut wiring up the component.

RusGraf:
> Testing isn't in scope for this ticket; this is just wiring up the component.

Ok, then how would you validate that the code works? Perhaps the cards split is not ideal to make it demo-able and testable.

ldk/javascript/src/whisper/mapper.ts
RusGraf: There is a 'path: string;' which is returned. Perhaps we could set this in the sitemap? Was wondering if we are going to send no value back with the stateMap to the loop author, then our direction on Form Submit is not going to be correct, as they wont be able to validate it. Thoughts?

richseviora: These are great questions and ideas...  I will spend more time thinking about this... as part of HELPS-854.

richseviora:
I've updated this PR to point to a feature branch I've established. So we won't need to get the tests working immediately, but in the next card (854) we should be able to add some tests.

RusGraf (approved):

epiccoleman (approved):

PR #278, opened 2021-07-15
HELPS-788: Omit unnecessary claims from User Aptitude JWT

epiccoleman:
This commit adds an optional includeEmail parameter to the User Aptitude
jwt function. When this parameter is set to true, the JWT will include
the current user's email address. When the parameter is unset, or false,
the email address will not be included in the JWT's claims, limiting
potential PII exposure.

### DIFF
--- a/docs/src/markdown-pages/aptitudes/user.createJWT.md
+++ b/docs/src/markdown-pages/aptitudes/user.createJWT.md
@@ -2,7 +2,7 @@
 name: "JWT"
 links_js: "jwt"
 ---
-Receive a [JWT](https://jwt.io) identifying the currently logged in Olive Helps user.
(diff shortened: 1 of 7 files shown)


richseviora:
Immediate thought - I think it's fairly likely that we'll add other optional claims. So instead of making this optional parameter a boolean value, could we instead look at an optional object with an email property?

epiccoleman:
> Immediate thought - I think it's fairly likely that we'll add other optional claims. So instead of making this optional parameter a boolean value, could we instead look at an optional object with an email property?

Yes! Good idea. Should be straightforward enough. I'll take a look this afternoon.

RusGraf (approved):

DanielWagersOlive (approved):

PR #279, opened 2021-07-15
HELPS-837 Fix problem where Select, Radio, and Checkboxes do not track state after a whisper update.

BrettMugglinOlive:


ldk/javascript/examples/self-test-loop/src/tests/whisper/whisper-update.ts
```diff
@@ -377,3 +379,123 @@ export const testWhisperStateOnChange = (): Promise<boolean> =>
...
+      label: 'Updated Whisper label :o',
+      components: [checkbox, radioButton, selectInput, ...secondUpdate.button],
+    });
+    await secondUpdate.promise;
+    resolve(true);
+  });
+
+export const testNonTextInputsWithValue = (): Promise<boolean> =>
+  new Promise(async (resolve, reject) => {
+    const selectValue = 1;
```
RusGraf, line 443: you may want to wrap all await functions with try/catch to catch any exceptions and resolve/reject gracefully.

RusGraf (approved):

epiccoleman (approved):

PR #282, opened 2021-07-17
HELPS-456 Add tests for opening files and directories.

BrettMugglinOlive:


ldk/javascript/examples/self-test-loop/src/tests/filesystem/index.ts
```diff
@@ -293,3 +293,76 @@ export const testFileExists = (): Promise<boolean> =>
...
     });
   });
+
+export const testOpenFile = (): Promise<boolean> =>
+  new Promise(async (resolve, reject) => {
+    const filePath = `test.txt`;
+    const writeMode = 0o755;
+
+    network
+      .encode('some text')
```
RusGraf, line 303: No need to encode value. filesystem.write() takes string as well.

ldk/javascript/examples/self-test-loop/src/tests/filesystem/index.ts
```diff
@@ -293,3 +293,76 @@ export const testFileExists = (): Promise<boolean> =>
...
+      .then((encodedValue) => {
+        setTimeout(() => {
+          filesystem
+            .writeFile({
+              path: filePath,
+              data: encodedValue,
+              writeOperation: filesystem.WriteOperation.overwrite,
+              writeMode,
+            })
+            .then(() => {
```
RusGraf, line 313: I'm actually rebuilding filesystem test group in await async pattern right now. Perhaps it would makes sense to drop the act().then().catch() pattern here to be aligned.

ldk/javascript/examples/self-test-loop/src/tests/filesystem/index.ts
```diff
@@ -293,3 +293,76 @@ export const testFileExists = (): Promise<boolean> =>
...
+                    .remove(filePath)
+                    .then(() => {
+                      setTimeout(() => {
+                        resolve(true);
+                      }, 1500);
+                    })
+                    .catch((error) => {
+                      setTimeout(() => {
+                        reject(error);
+                      }, 1500);
```
RusGraf, line 326: Do you need to wait for a rejection? Perhaps I'm not understanding the value.

ldk/javascript/examples/self-test-loop/src/tests/filesystem/index.ts
```diff
@@ -293,3 +293,54 @@ export const testFileExists = (): Promise<boolean> =>
...
+
+    try {
+      await filesystem.writeFile({
+        path: filePath,
+        data: 'some text',
+        writeOperation: filesystem.WriteOperation.overwrite,
+        writeMode,
+      });
+      await filesystem.openWithDefaultApplication(filePath);
+      sleep(2000);
```
RusGraf, line 310: why not use setTimeout() since we already have this concept in the selfTestLoop?

RusGraf (approved):

ldk/javascript/examples/self-test-loop/src/config/filesystem-group.ts
```diff
@@ -47,4 +47,22 @@ export const filesystemTestGroup = (): TestGroup =>
...
+      'File Aptitude - Open Directory',
+      filesystemTests.testOpenDirectory,
+      10000,
+      'Attempting to open directory...',
+    ),
+    new LoopTest(
+      'File Aptitude - Open File Does Not Exists',
+      filesystemTests.testOpenFileDoesNotExist,
+      10000,
+      'Trying to open afile that does not exist',
```
cshields-oliveai, line 66: "afile"

flowermm (approved):

ldk/javascript/examples/self-test-loop/src/config/filesystem-group.ts
MaggieMaSS: Would it be ``` 'Trying to open a file that does not exist',```

PR #291, opened 2021-07-21
Helps 917 - making sure close handler is called when websocket close is triggered.

RusGraf:


ldk/javascript/examples/self-test-loop/src/tests/network/index.ts
```diff
@@ -117,49 +116,47 @@ export const testWebsocketConnection = (): Promise<boolean> =>
...
         if (error) {
-          console.error(`OnCloseHandler received error:`);
           console.error(error);
-
-          return;
+          reject(error);
         }
-
         console.info(`Received on close code: ${code}. ${text}`);
+        if (textTestPassed && binaryTestPassed) {
```
MaggieMaSS, line 128: `textTestPassed` and `binaryTestPassed` are false here, may I know why we want it to resolve true with false `textTestPassed` and false `binaryTestPassed`

RusGraf: Well, this is a callback scope. So the idea is to only resolve test as passed if we already received 2 messages: text and binary, before we get into this callback.

flowermm (approved):

BrettMugglinOlive (approved):

PR #292, opened 2021-07-21
HELPS-877: copyable messages

cshields-oliveai:


ldk/javascript/examples/self-test-loop/src/tests/whisper/index.ts
```diff
@@ -430,6 +432,110 @@ export const testListPairWithCopyableLabel = (): Promise<boolean> =>
...
+        {
+          type: WhisperComponentType.Message,
+          copyable: MessageWhisperCopyMode.Body,
+          body: copyableText,
+        },
+      ],
+    });
+
+    setTimeout(async () => {
+      const response = await clipboard.read();
```
RusGraf, line 492: why not use clipboard.listen() instead. So you could avoid a set 5 sec delay?

cshields-oliveai: Good idea.

cshields-oliveai: Currently, the listen callback is not invoked when the electron app copies values to the clipboard. We will save this for a future user story if it is an issue.

RusGraf (approved):

epiccoleman (approved):

PR #293, opened 2021-07-22
HELPS-937 Add the AlignItems property to Box components.

BrettMugglinOlive:
In addition to that
- addressed tech debt around how the JustifyContent names were being used (added FlexStart and FlexEnd). 
- Alphabetized. 
- Changed all incorrect JustifyContent properties to something that actually corresponded with a property. The default property was "SpaceBetween", and "SpaceAround" and "SpaceEvenly" weren't actually doing anything.

### DIFF
--- a/ldk/javascript/src/whisper/mapper.ts
+++ b/ldk/javascript/src/whisper/mapper.ts
@@ -32,9 +32,11 @@ export function mapToInternalChildComponent(
     case WhisperComponentType.Box:
       // eslint-disable-next-line
       const { onClick } = component;
+      console.log(component.alignItems);
       if (onClick) {
         return {
           id: component.id,
+          alignItems: component.alignItems,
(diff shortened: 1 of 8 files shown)


ldk/javascript/src/whisper/mapper.ts
```diff
@@ -32,9 +32,11 @@ export function mapToInternalChildComponent(
...
       // eslint-disable-next-line
       const { onClick } = component;
+      console.log(component.alignItems);
       if (onClick) {
         return {
           id: component.id,
+          alignItems: component.alignItems,
           alignment: 'justifyContent' in component ? component.justifyContent : component.alignment,
           direction: component.direction,
           key: component.key,
```
RusGraf, line 42: Would `...component` work here, so we don't have to map identical params?

RusGraf (approved):

epiccoleman (approved):

cshields-oliveai (approved):

PR #295, opened 2021-07-22
Helps 832

flowermm:
Note that LayoutOptions will continue to expand as users request more styling like padding, margins, widths, and heights. Leaving it at "flex" only for this PR to get the architecture in place on the Sidekick side of things.


ldk/javascript/src/whisper/mapper.ts
```diff
@@ -218,35 +199,20 @@ export function mapToInternalComponent(
...
-              mapToInternalChildComponent(childComponent, stateMap),
-            ),
-          ),
-          type: WhisperComponentType.CollapseBox,
-          onClick: (error: Error, param: boolean, whisper: OliveHelps.Whisper) => {
-            onClick(error, param, mapToExternalWhisper(whisper, stateMap));
-          },
-        } as OliveHelps.CollapseBox;
-      }
+      const onClick = 'onClick' in component ? component.onClick : null;
```
flowermm, line 202: Simplified these because sidekick doesn't care if the property is undefined when it receives it.

ldk/javascript/src/whisper/types.ts
```diff
@@ -257,6 +262,7 @@ export type Message = WhisperComponent<WhisperComponentType.Message> & {
 export type Divider = WhisperComponent<WhisperComponentType.Divider>;
 
 export type ChildComponents =
+  | Box
```
flowermm, line 265: Fixed a bug where Box was not listed as a child component, even though it was listed that way in sidekick

RusGraf: This was done on a purpose. Only Box and Deprecated box should allow box in box capability. After your change you added this behavior to the CollapseBox as well.

ldk/javascript/src/whisper/mapper.ts
```diff
@@ -28,39 +29,24 @@ export function mapToInternalChildComponent(
   component: BoxChildComponent,
   stateMap: StateMap,
 ): OliveHelps.ChildComponents {
+  const onClick = 'onClick' in component ? component.onClick : null;
   switch (component.type) {
     case WhisperComponentType.Box:
-      // eslint-disable-next-line
-      const { onClick } = component;
```
RusGraf, line 34: image

ldk/javascript/src/whisper/types.ts
flowermm: Our internal test code then allows for two different behaviors - I cannot list Box as a child of CollapseBox in our testing suite. However, I can do so in my own personal loop (and it works). This is due to two different abilities in types.ts here vs. the index.d.ts file.  I was trying to remedy the two different allowed behaviors (In index.d.ts Box is listed as allowed for CollapseBox) So, which do we want? 😅  Happy to change whichever file (ie remove Box in the index.d.ts file instead of adding it to this one)

flowermm: Also this brings up the amount of copy paste we have and why it could be dangerous. I wonder if there is an easier way to sync the two files.

RusGraf: Perhaps we could chat about it real quick. I just want to indicate that you changing the existing feature behavior with this change. Perhaps we need to run it passed the PO to make sure we aligned. But in the past it was described to me that only box and not a collapse box could support box in the box feature.

flowermm: ldk/javascript/src/types/oliveHelps/index.d.ts is the other file I'm referring too by the way in case anyone is following this thread. It does expose box as a child component, so users can already access it (the update in this file only helps our testing suite). After following up with Ruslan, we are asking Robin which is the correct behavior so I can fix it 👍  The other files of interest are collapseBox and Box in sidekick - collapseBox already has the ability to hold a box as a child.  Box does NOT support collapse box as a child however, and will alert the user that it is not valid. This change does not impact that implementation.

RusGraf (approved):

ldk/javascript/src/whisper/mapper.ts
cshields-oliveai: cool

ldk/javascript/src/whisper/types.ts
flowermm: Okay - Robin is fine with CollapseBox being able to have a Box since sidekick and the exposed namespace already support it (whether it was intended or not ha). We are good to go here 😄

cshields-oliveai: I do not love the duplication between `ldk/javascript/src/types/oliveHelps/index.d.ts` and individual aptitude `types.ts` files either, but it _does_ also allow us to provide easier to use APIs for Loop authors, and then translate into something else before sending requests over to OH. The intent is that Loop Authors only get a hold of types defined by related `aptitudeX/types.ts` files. We've been differentiating the two by calling them 'internal' and 'external' - where 'internal' is only the contract between LDK and OH.

How to reduce duplication from a technical sense - I am not sure.

cshields-oliveai (approved):

PR #300, opened 2021-07-29
[Helps 184] Add Search Aptitude Service

alexmatthew-olive:
Added Search Aptitude with a few unit test for the search interface.

Note: On the `SearchResult` type there is still some research to be done to make sure we are constraining the key types to string. 

Please let me know if you have any questions


ldk/javascript/src/search/index.ts
```diff
@@ -0,0 +1,90 @@
...
+  search(term: string): Promise<SearchResult>;
+  /**
+   * Executes a search with the using a query string to search over the data.
+   *
+   * @param queryString - The query string that will be used to search over the index.
+   * @returns A Promise resolving with the search results as a string
+   */
+  queryStringSearch(queryString: string): Promise<SearchResult>;
+  /**
+   * Executes a search with the using a query string to search over the data.
```
cshields-oliveai, line 26: I believe this descriptions for `queryStringSearch` and `update` may be typos.

"Executes a search with the using a query string to search over the data."

ldk/javascript/src/search/index.ts
```diff
@@ -0,0 +1,90 @@
...
+   *
+   * @param documents - The date that will be indexed and searchable
+   * @param config - The configuration object defines how the index will preform certain searches
+   */
+  update(documents: Document[], config: Config): Promise<void>;
```
cshields-oliveai, line 31: It is a little confusing to me how `Document[]` maps to some kind of date. Do you mean "data"?

alexmatthew-olive: Should of been `data`

ldk/javascript/src/search/mapper.ts
```diff
@@ -0,0 +1,104 @@
...
+  name: field.name,
+  displayName: field.displayName,
+  type: mapToFieldType(field.type),
+});
+
+export const mapToFields = (fields: Field[]): OliveHelps.Field[] =>
+  fields.map((f) => mapToField(f));
+
+export const mapToSearchResult = (searchResult: OliveHelps.SearchResult): SearchResult => ({
+  data: searchResult.data,
```
cshields-oliveai, line 21: I believe that for types that have **no difference** between their internal and external representations, you just spread the incoming object out into the new type:

For example (something like),

```typescript
export const mapToSearchResult = (searchResult: OliveHelps.SearchResult): SearchResult => ({
    ...searchResult
});
```

ldk/javascript/src/search/search.test.ts
```diff
@@ -0,0 +1,57 @@
...
+      const config: search.Config = { searchSize: 10 };
+      search.openIndex(indexName, config);
+      return expect(oliveHelps.search.openIndex).toBeCalledWith(
+        indexName,
+        config,
+        expect.any(Function),
+      );
+    });
+
+    it('exists returns a promise result with expected index', () => {
```
cshields-oliveai, line 51: small nitpick with test names, usually you want to add a describe block for each noun, and the `it()` functions inside describe a behavior that noun has. So for this test it could be something like:

```
describe('exists', () => {
    it('returns a promise result with expected index', () => { ... });
});
```

ldk/javascript/src/webpack/generate-banner.test.ts
```diff
@@ -27,6 +27,7 @@ describe('Generate Banner', () => {
         keyboard: {},
         network: { urlDomains: [{ value: '*.google.com' }] },
         process: {},
+        search: {},
```
cshields-oliveai, line 30: You probably know this better than me, but I believe a change in metadata in the LDK will also require a schema change within LoopLibrary.

ldk/javascript/src/search/types.ts
```diff
@@ -0,0 +1,86 @@
...
+	exactMatchThreshold?: number;
+	/**
+   * Add the ability to do a prefix search along with the fuzzy search. Default is true.
+   */
+	beginsWithSearch?: boolean;
+}
+
+export interface SearchResult {
+	/**
+   * Search result data comes in as an array of objects
```
cshields-oliveai, line 79: Should the meaning of the `key` and it's associated values also be described in the comment?

ldk/javascript/src/webpack/generate-banner.test.ts
alexmatthew-olive: Yes, you are correct, that is on the list of things we are planning on still getting done

ldk/javascript/src/search/types.ts
alexmatthew-olive: Only left that out because as I was talking to Rich about this and we were discussing how to possible constrain the key type to string so there will be some changes and comment changes coming to this but thanks for noting this.

ldk/javascript/src/search/types.ts
```diff
@@ -0,0 +1,86 @@
...
+  type?: FieldType;
+};
+
+export type Document = {
+  /**
+   * The name of the document
+   */
+  name: string;
+  /**
+   * The data for the document that will be indexed and searched over
```
BrettMugglinOlive, line 49: Is there a specific format the data needs to be in in order to be indexed?

ldk/javascript/src/search/types.ts
```diff
@@ -0,0 +1,86 @@
+export enum FieldType {
```
BrettMugglinOlive, line 1: This enum is called `FieldType`, but the comments refer to a `Filter`. Is that intentional?

BrettMugglinOlive (commented):
I'm not sure who should take responsibility for this, but given this is a new feature, and it's using a lot of different terminology within itself, it feels like our developer hub needs to be updated to give an explanation of how all these different parts interact.

ldk/javascript/src/search/types.ts
alexmatthew-olive: > I'm not sure who should take responsibility for this, but given this is a new feature, and it's using a lot of different terminology within itself, it feels like our developer hub needs to be updated to give an explanation of how all these different parts interact.

To answer your first question, I have some written documentation but  I may need to pair with someone who has written some documentation for the LDK so it can be consistent with that is there

The data needs to be in a json like syntax, I have that in the documentation I have written but probably should add it in the comments here. It can accept valid json, typed object, or an array of typed objects that have been called with `JSON.stringify(...)`

ldk/javascript/src/search/types.ts
alexmatthew-olive: The `FieldType` is used to describe how we want the field to be interpreted and what we want inserted into the index. The field type mapping themselves apply filters (but do other things as well, like reflection and have specific analyzer for specific data types. The reason I put the filters in the comments is because some of the mappings will filter out certain data (ex english stop words). But we could revisit the naming scheme though, I do see how it can be a bit confusing

ldk/javascript/src/search/index.ts
```diff
@@ -0,0 +1,90 @@
...
+  promisifyMappedWithTwoParams,
+  promisifyMappedWithThreeParams,
+  promisifyWithParam,
+} from '../promisify';
+import * as mapper from './mapper';
+import { Document, Config, SearchResult } from './types';
+
+export * from './types';
+
+export interface Index {
```
RusGraf, line 11: Should this be moved to `./types.ts`?

ldk/javascript/src/search/mapper.ts
```diff
@@ -0,0 +1,100 @@
...
+export const mapToField = (field: Field): OliveHelps.Field => ({
+  name: field.name,
+  displayName: field.displayName,
+  type: mapToFieldType(field.type),
+});
+
+export const mapToFields = (fields: Field[]): OliveHelps.Field[] =>
+  fields.map((f) => mapToField(f));
+
+export const mapToSearchResult = (searchResult: OliveHelps.SearchResult): SearchResult => ({
```
RusGraf, line 20: Is this being used anywhere?

ldk/javascript/src/search/mapper.ts
```diff
@@ -0,0 +1,100 @@
...
+          },
+        );
+      } catch (e) {
+        handleCaughtError(reject, e, 'queryStringSearch');
+      }
+    }),
+  update: (documents: Document[], config: Config): Promise<void> =>
+    new Promise<void>((resolve, reject) => {
+      try {
+        index.update(mapToDocuments(documents), config, (error: Error | undefined) => {
```
RusGraf, line 75: I'm wondering if you could just do: `index.update(documents, config, (error: Error | undefined) => {` and eliminate most of the mappers. If 2 objects have same fields they will convert by default, no special mapping needed. I believe this will also cleanup your promisify methods as well.

ldk/javascript/src/types/oliveHelps/index.d.ts
```diff
@@ -201,6 +222,55 @@ declare namespace OliveHelps {
...
+      queryString: string,
+      callback: (error: Error | undefined, searchResult: SearchResult) => void,
+    ): void;
+    update(
+      documents: Array<Document>,
+      config: Config,
+      callback: (error: Error | undefined) => void,
+    ): void;
+    delete(callback: (error: Error | undefined) => void): void;
+  }
```
RusGraf, line 247: I think this should work for you as well: 
`interface Index {
    search: ReadableWithParam<string, SearchResult>;
    queryStringSearch: ReadableWithParam<string, SearchResult>;
    update: ReadableWithTwoParams<Array<Document>, Config, void>;
    delete: Readable<void>;
  }`

ldk/javascript/src/search/mapper.ts
```diff
@@ -0,0 +1,100 @@
...
+  ...config,
+});
+
+export const mapToIndex = (index: OliveHelps.Index): Index => ({
+  search: (term: string): Promise<SearchResult> =>
+    new Promise<SearchResult>((resolve, reject) => {
+      try {
+        index.search(term, (error: Error | undefined, searchResult: SearchResult) => {
+          if (error) {
+            console.error(`Received error on result: ${error.message}`);
```
RusGraf, line 46: It would make sense to handle error in here the same way as you did in the catch block. Since you already abstracted the handler:
`handleCaughtError(reject, error as Error, 'search');`
Same goes for the rest of the methods.

ldk/javascript/src/search/index.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export function createIndex(
+  name: string,
+  documents: Array<Document>,
+  config: Config,
+): Promise<Index> {
+  return promisifyMappedWithThreeParams(
+    name,
+    documents,
+    mapper.mapToDocuments,
```
RusGraf, line 72: You could avoid extra mappings as your params are the same and simply do:
`
export function createIndex(
  name: string,
  documents: Array<Document>,
  config: Config,
): Promise<Index> {
  return promisifyMappedWithThreeParams(
    name,
    documents,
    config,
    mapper.mapToIndex,
    oliveHelps.search.createIndex,
  );
}

export function openIndex(name: string, config: Config): Promise<Index> {
  return promisifyMappedWithTwoParams(
    name,
    config,
    mapper.mapToIndex,
    oliveHelps.search.openIndex,
  );
}
`

Then in promisify just remove mappers, like:
`
export function promisifyMappedWithTwoParams<
  TParam1,
  TParam2,
  TInternalOut,
  TExternalOut
>(
  param1: TParam1,
  param2: TParam2,
  map: Mapper<TInternalOut, TExternalOut>,
  arg: OliveHelps.ReadableWithTwoParams<TParam1, TParam2, TInternalOut>,
): Promise<TExternalOut> {
  return new Promise((resolve, reject) => {
    try {
      arg(param1, param2, (error, value) => {
        if (error) {
          reject(error);
          return;
        }
        resolve(map(value));
      });
    } catch (e) {
      handleCaughtError(reject, e as Error);
    }
  });
}

export function promisifyMappedWithThreeParams<
  TParam1,
  TParam2,
  TParam3,
  TInternalOut,
  TExternalOut
>(
  param1: TParam1,
  param2: TParam2,
  param3: TParam3,
  map: Mapper<TInternalOut, TExternalOut>,
  arg: OliveHelps.ReadableWithThreeParams<TParam1, TParam2, TParam3, TInternalOut>,
): Promise<TExternalOut> {
  return new Promise((resolve, reject) => {
    try {
      arg(param1, param2, param3, (error, value) => {
        if (error) {
          reject(error);
          return;
        }
        resolve(map(value));
      });
    } catch (e) {
      handleCaughtError(reject, e as Error);
    }
  });
}
`

The only thing you would have to map is the return types.

ldk/javascript/src/search/index.ts
alexmatthew-olive: My thought for this was `Index` and `Search` are interfaces with methods only and the other types were interfaces with properties only, but I can move it if you think it makes more sense there.

ldk/javascript/src/search/mapper.ts
```diff
@@ -0,0 +1,100 @@
...
+  fields: document.fields ? mapToFields(document.fields) : undefined,
+});
+
+export const mapToDocuments = (documents: Document[]): OliveHelps.Document[] => {
+  const docs: OliveHelps.Document[] = [];
+  documents.forEach((d) => docs.push(mapToDocument(d)));
+  return docs;
+};
+
+export const mapToConfig = (config: Config): OliveHelps.Config => ({
```
richseviora: Hmm - this shouldn't be necessary unless there's a specific need to duplicate the object, otherwise these types are compatible.

I think you said you're trying to follow the pattern we established elsewhere (for example in the Whisper mapping logic), and that may not be necessary here if the types are pretty similar.

ldk/javascript/src/search/mapper.ts
```diff
@@ -0,0 +1,100 @@
...
+export const mapToDocument = (document: Document): OliveHelps.Document => ({
+  name: document.name,
+  data: document.data,
+  fields: document.fields ? mapToFields(document.fields) : undefined,
+});
+
+export const mapToDocuments = (documents: Document[]): OliveHelps.Document[] => {
+  const docs: OliveHelps.Document[] = [];
+  documents.forEach((d) => docs.push(mapToDocument(d)));
+  return docs;
```
richseviora: JS makes mapping a lot easier :) 

```suggestion
  return documents.map(mapToDocument);
```

ldk/javascript/src/search/mapper.ts
```diff
@@ -0,0 +1,100 @@
...
+};
+
+export const mapToConfig = (config: Config): OliveHelps.Config => ({
+  ...config,
+});
+
+export const mapToIndex = (index: OliveHelps.Index): Index => ({
+  search: (term: string): Promise<SearchResult> =>
+    new Promise<SearchResult>((resolve, reject) => {
```
richseviora, line 11: I'm wondering if it'd be better to turn this into a class. I'll pull this down and noodle over that.

ldk/javascript/src/search/mapper.ts
```diff
@@ -0,0 +1,100 @@
...
+  update: (documents: Document[], config: Config): Promise<void> =>
+    new Promise<void>((resolve, reject) => {
+      try {
+        index.update(mapToDocuments(documents), config, (error: Error | undefined) => {
+          if (error) {
+            console.error(`Received error on result: ${error.message}`);
+            reject(error);
+          }
+        });
+        resolve();
```
richseviora, line 50: This resolve call should be moved to within the `index.update` callback; right now it'll fire off the update call and then move onto the resolve without waiting for the result to necessarily return.

ldk/javascript/src/search/mapper.ts
```diff
@@ -0,0 +1,100 @@
...
+  delete: (): Promise<void> =>
+    new Promise<void>((resolve, reject) => {
+      try {
+        index.delete((error: Error | undefined) => {
+          if (error) {
+            console.error(`Received error on result: ${error.message}`);
+            reject(error);
+          }
+        });
+        resolve();
```
richseviora, line 64: Also move this one up.

ldk/javascript/src/promisify.ts
```diff
@@ -30,6 +30,54 @@ export function promisifyMappedBothWithParams<TParamIn, TParamOut, TInternalOut,
   });
 }
 
+export function promisifyMappedWithTwoParams<TParam1, TParam2, TInternalOut, TExternalOut>(
+  param1: TParam1,
+  param2: TParam2,
+  map: Mapper<TInternalOut, TExternalOut>,
+  arg: OliveHelps.ReadableWithTwoParams<TParam1, TParam2, TInternalOut>,
+): Promise<TExternalOut> {
+  return new Promise((resolve, reject) => {
```
richseviora, line 39: I really need to spend some time looking into these types and figure out if there's an easier way to express all these functions.

richseviora (changes requested):

ldk/javascript/src/types/oliveHelps/index.d.ts
alexmatthew-olive: should be completed

ldk/javascript/src/search/index.ts
RusGraf: I see your point. I don't think we ever clearly defined what goes where as a team. I usually just leave an interface which describes the top level of the aptitude and place rest in the types. But I don't see big issue here regardless.

RusGraf (approved):

ldk/javascript/src/search/mapper.ts
richseviora: This is a future refactoring exercise. Don't need to address this right now.

richseviora (approved):

alexmatthew-olive:
Bumped LDK Contract Version please let me know if I missed it anywhere else

PR #302, opened 2021-07-30
HELPS-975: Add Icon Component

cshields-oliveai:
Add a new Icon component to the LDK:

```typescript
type Icon = Component<'icon'> & {
    name: string;
    size?: IconSize;
    color?: Color;
    onClick?: WhisperHandler;
    tooltip?: string;
  };
```


ldk/javascript/src/types/oliveHelps/index.d.ts
```diff
@@ -340,12 +341,14 @@ declare namespace OliveHelps {
 
   type DateTimeType = 'date' | 'time' | 'date_time';
 
+  type IconSize = 'small' | 'medium' | 'large' | 'x-large';
+
+  type Color = 'white' | 'grey' | 'black' | 'whisper-strip';
```
cshields-oliveai, line 346: Note that `black` and `whisper-strip` are not possibilities of the `SectionTitle` component. In those cases, if those colors a chosen by Loop Author it will currently fallback to its default color. This Color property is TBD

richseviora: Just to confirm - that is just an interim implementation until we figure out what changes we want to make to our approach to colour, right? Otherwise I really don't think we should support someone assigning invalid values to a component.

RusGraf: So if you planing to add extra colors to the SectionTitle component, do you need to add mapping for them in the Sidekick? If not then all unmapped values will be defaulted to 'white'.

ldk/javascript/src/whisper/mapper.ts
```diff
@@ -188,6 +188,18 @@ export function mapToInternalChildComponent(
           component.onChange(error, param, mapToExternalWhisper(whisper, stateMap));
         },
       } as OliveHelps.DateTimeInput;
+    case WhisperComponentType.Icon:
+      // eslint-disable-next-line
```
RusGraf, line 192: What do you think if we just disable a specific rule for a whole file here?

ldk/javascript/src/types/oliveHelps/index.d.ts
cshields-oliveai: `white` is currently the default value for `SectionTitle`. We will discuss this with UX.

ldk/javascript/src/whisper/mapper.ts
cshields-oliveai: According to the docs for "no-case-declarations", eslint can be satisfied here (with its default config) if we surround cases which declare variables with {} - so Icon could become:

```typescript
case WhisperComponentType.Icon: {
      const onIconClick = component.onClick;
      if (onIconClick) {
        return {
          ...component,
          onClick: (error, whisper) => {
            onIconClick(error, mapToExternalWhisper(whisper, stateMap));
          },
        } as OliveHelps.Icon;
      }
      return component as OliveHelps.Icon;
    }
```

What does the team think of this?

RusGraf: Interesting, I have not thought of this. But here is another way to handle it. It doesn't have to be part of your card Cory. Use a map instead of the switch/case. Map<WhisperComponentType, ComponentHandler>. Then trigger handler map.get(component.type).execute(); In this case handler logic will be less coupled and you could abstract it to another file if needed.

cshields-oliveai: I got you. Lets work on a refactor for this as part of another card.

RusGraf (approved):

ldk/javascript/src/types/oliveHelps/index.d.ts
```diff
@@ -313,8 +313,9 @@ declare namespace OliveHelps {
     | 'select'
     | 'telephone'
     | 'textInput'
-    | 'sectionTitle'
-    | 'dateTimeInput';
+    | 'dateTimeInput'
+    | 'icon'
+    | 'sectionTitle';
```
flowermm, line 318: Might as well alphabetize these for posterity / ease of reference between this file and types.js

flowermm (approved):

PR #308, opened 2021-08-09
[HELPS-856] Add file reading support

richseviora:
This PR adds support for reading files from the File object directly. 

Some cleanup todos:

- [x] Update the documentation about how to use these objects and their limitations.
- [x] Look at mapping again and see if I can refactor that a bit more cleanly.


ldk/javascript/src/whisper/types.ts
```diff
@@ -238,13 +240,35 @@ export type Message = WhisperComponent<WhisperComponentType.Message> & {
...
   onDrop: WhisperHandlerWithParam<File[]>;
   tooltip?: string;
   validationError?: string;
+  /**
+   * The value property can be used to update the list of files selected, including dropping them. If the value property
+   * is null or undefined, the existing selection will remain as-is. To deselect all the files, set it to an empty
+   * array.
+   *
+   * You cannot add a file to the selection via this interface. The component will ignore any files provided it
+   * currently does not have selected.
```
cshields-oliveai, line 270: "currently does not have selected." Is this WIP? Confused by that last bit.

ldk/javascript/src/whisper/types.ts
```diff
@@ -238,13 +240,35 @@ export type Message = WhisperComponent<WhisperComponentType.Message> & {
 };
 
 export type DropZone = WhisperComponent<WhisperComponentType.DropZone> & {
+  /**
+   * Specify what extensions the user can include. Do not include the dot.
+   *
+   * @example ['jpg','jpeg','png']
```
cshields-oliveai, line 246: Do we want to limit the file extension types that can be sent inside of the LDK?

ldk/javascript/src/whisper/types.ts
richseviora: It's intended to be final; If the Loop author provides a File object they had stored from a previous event but that file has since been deselected, that file object will be ignored.

cshields-oliveai: Got it.

ldk/javascript/src/whisper/types.ts
richseviora: I don't think we'd want to create an allowlist inside the LDK itself? It could be overrode if they accessed the internal `oliveHelps` object and the author isn't required to provide an extension list. + we don't know what extensions a Loop may want to work with.

cshields-oliveai (approved):

ldk/javascript/src/whisper/mapper.ts
```diff
@@ -116,7 +117,24 @@ export function mapToInternalChildComponent(
...
         ...component,
         onDrop: (error, param, whisper) => {
-          component.onDrop(error, param, mapToExternalWhisper(whisper, stateMap));
+          const callbackHandler: (file: OliveHelps.File) => File = (file: OliveHelps.File) => ({
+            path: file.path,
+            size: file.size,
+            readFile: () =>
+              new Promise<Uint8Array>((resolve, reject) => {
+                file.readFile((readError, buffer) => {
+                  if (error) {
```
RusGraf, line 126: Should this be `if (readError) {` ?

RusGraf (approved):

ldk/javascript/src/whisper/mapper.ts
richseviora: Yes! Great catch, thanks!

ldk/javascript/src/whisper/types.ts
flowermm: _The component will ignore any provided files that are not currently selected_

ldk/javascript/src/whisper/types.ts
```diff
@@ -238,13 +240,35 @@ export type Message = WhisperComponent<WhisperComponentType.Message> & {
...
+   */
   noun?: string;
+  /**
+   * The callback function to call whenever the user selects or unselects a file.
+   */
   onDrop: WhisperHandlerWithParam<File[]>;
   tooltip?: string;
   validationError?: string;
+  /**
+   * The value property can be used to update the list of files selected, including dropping them. If the value property
```
flowermm, line 265: You can probably just write the first sentence as: _The value property can be used to modify the selected files._ For some reason "including dropping them" is bending my mind (you mean remove right? I haven't had coffee yet, send help). Anyway, the next few sentences clearly state what you mean :)

flowermm (approved):

PR #313, opened 2021-08-12
Exclude email self test

epiccoleman:
This commit adds a test to the self test loop for exercising the JWT call when called with `{includeEmail: false}`.

### DIFF
--- a/ldk/javascript/examples/self-test-loop/package.json
+++ b/ldk/javascript/examples/self-test-loop/package.json
@@ -33,7 +33,8 @@
       "filesystem": {},
       "process": {},
       "ui": {},
-      "user": {},
+      "user": {
+      },
       "system": {},
       "keyboard": {},
       "whisper": {},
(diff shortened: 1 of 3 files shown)


ldk/javascript/examples/self-test-loop/package.json
```diff
@@ -33,7 +33,8 @@
       "filesystem": {},
       "process": {},
       "ui": {},
-      "user": {},
+      "user": {
```
DanielWagersOlive, line 36: 🤔

epiccoleman: Whoops. This needs rebased anyway - gimme a minute to wrangle some history around.

DanielWagersOlive (approved):

ldk/javascript/examples/self-test-loop/src/tests/user/index.ts
```diff
@@ -5,7 +5,40 @@ export const testJwt = (): Promise<boolean> =>
     user.jwt().then((token) => {
       if (token) {
         whisper.create({
-          label: 'User Aptitude JWT - No Email',
+          label: 'User Aptitude JWT - Default Claims',
```
epiccoleman, line 8: the diff here is a bit confusing. the change might make a little more sense if you view the whole file. This top test just had its name changed. The new test is below, and is basically the same test, except that `user.jwt` is called with the config parameter, like such as: `user.jwt({includeEmail: false})`

RusGraf (approved):

flowermm (approved):

PR #316, opened 2021-08-19
Helps 781

flowermm:


ldk/javascript/examples/self-test-loop/src/tests/whisper/index.ts
```diff
@@ -1802,3 +1802,227 @@ export const testFlex = (): Promise<boolean> =>
...
+                      flex: '1',
+                    },
+                    onClick: () => console.log('Click the Button'),
+                  },
+                  {
+                    type: whisper.WhisperComponentType.Icon,
+                    name: 'favorite',
+                    size: whisper.IconSize.Small,
+                    layout: {
+                      flex: '1',
```
MaggieMaSS, line 1842: That might not be related to this task, but do you think that's a better idea to make flex an enum?

flowermm: Unfortunately, Flex can take any valid flex string. So things like "1 0 50%", or "2", or even "0 1".  This property is def for power users. I think we could consider some defaults / enums, but we weren't sure how to go about doing that (we being Rich, JB, and I). Open to ideas!

cshields-oliveai (approved):

PR #317, opened 2021-08-20
Helps 1148 excel aptitude

MaggieMaSS:
Pull Request of Helps-1148 Excel Aptitude

- Built Excel Aptitude within the LDK
- Provided unit tests for this  Aptitude
- Added self test loop

Since sidekick uses xlsx as the name, so I use xlsx instead of excel here to keep the consistency.


README.md
```diff
@@ -1,3 +1,22 @@
-# Loop Development Kit
```
cshields-oliveai, line 1: What happened with these readme files? Did these already exist and get moved? Was this part of your merge? I assume you did not add these as part of this excel PR?

MaggieMaSS: @cshields-oliveai This is not supposed to be my part of change. It seems to be @flowermm's Gitbook changes. these changes seems to be moved but I need to double check.

RusGraf (approved):

PR #328, opened 2021-08-31
[HELPS-981] Add React Whispers

richseviora:
I traded my soul for some React. Remember React? It's back, but in Whisper form.

This PR:

- Adds an implementation of the React host configuration to support creating and updating Whispers.
- Adds some high level usage instructions and a working example.
- Removes the Jest mocking of the `console` object and makes our tests silent by default.


ldk/javascript/src/whisper/react/renderer-config.ts
```diff
@@ -0,0 +1,330 @@
...
+export const config: CoreConfig & PersistenceConfig = {
+  afterActiveInstanceBlur: undefined,
+  akeOpaqueHydratingObject: undefined,
+  appendChildToContainerChildSet(childSet: ChildSet, child: Instance | TextInstance): ChildSet {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any -- just need to check type that doesn't exist in Instance
+    if ((child as any).type !== 'whisper') {
+      throw new Error('oh-whisper must be top-level element');
+    }
+    // In this function we are adding the Whisper properties to the ChildSet.
+    const whisper = (child as unknown) as NewWhisper;
```
nickharder88, line 151: Type predicates could be an efficient way to keep  type safety in a couple places here.

https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

ldk/javascript/src/whisper/react/renderer-config.ts
```diff
@@ -0,0 +1,330 @@
...
+  ): Instance {
+    const value: any = {
+      ...instance,
+      ...newProps,
+    };
+    delete value.children;
+    if (!keepChildren && value.components) {
+      value.components = [];
+    }
+    getHandlerByTagType(type).assignTextChildren?.(instance, newProps);
```
nickharder88, line 203: If we're cloning the instance, why are we assigning text children to the old instance and not to the "value"?

(I could be misunderstanding this function)

ldk/javascript/src/whisper/react/renderer-config.ts
```diff
@@ -0,0 +1,330 @@
...
+  isOpaqueHydratingObject(value: any): boolean {
+    return false;
+  },
+  isPrimaryRenderer: true,
+  makeClientId: undefined,
+  makeClientIdInDEV: undefined,
+  makeOpaqueHydratingObject: undefined,
+  noTimeout: undefined,
+  now(): number {
+    return 0;
```
nickharder88, line 292: Their documentation states - "You can proxy this to performance.now() or its equivalent in your environment.". Does returning 0 have any side effects?

ldk/javascript/src/whisper/react/whisper-render-instance.ts
```diff
@@ -0,0 +1,58 @@
...
+  private onCloseHandler: undefined | (() => Promise<void>);
+
+  async createOrUpdateWhisper(whisperData: NewWhisper | UpdateWhisper): Promise<void> {
+    if (this.status === RenderInstanceStatus.Closed) {
+      return;
+    }
+    if (this.whisper == null) {
+      await this.createWhisper(whisperData as NewWhisper);
+    } else {
+      if (this.status !== RenderInstanceStatus.Created) {
```
nickharder88, line 33: Redundant if layer? else if could be used

ldk/javascript/src/whisper/react/whisper-render-instance.ts
```diff
@@ -0,0 +1,58 @@
...
+export class WhisperRenderInstance implements WhisperRenderingInterface {
+  private status: RenderInstanceStatus = RenderInstanceStatus.NotCreated;
+
+  private whisper: Whisper | undefined;
+
+  private onCloseHandler: undefined | (() => Promise<void>);
+
+  async createOrUpdateWhisper(whisperData: NewWhisper | UpdateWhisper): Promise<void> {
+    if (this.status === RenderInstanceStatus.Closed) {
+      return;
```
nickharder88, line 28: Should this do something harsher than just a return? Below, if the status is in an invalid state then we throw an error. Would this be considered an invalid state?

ldk/javascript/package.json
```diff
@@ -35,7 +35,10 @@
     "url": "https://github.com/open-olive/loop-development-kit/issues"
   },
   "devDependencies": {
+    "@babel/plugin-transform-react-jsx": "^7.14.5",
     "@types/jest": "^26.0.22",
+    "@types/react": "^17.0.14",
+    "@types/react-reconciler": "^0.26.3",
```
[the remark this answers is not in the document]
richseviora: Great catch! I called that out in the readme but forgot to update that here. I think I'm actually going to pin it to the specific version until we get a better sense of what they're actually doing with versioning.

ldk/javascript/src/whisper/react/renderer-config.ts
richseviora: I'm a big fan of using type predicates, though I'm not 100% sure it makes sense in this particular context. Normally I'd use a type predicate to figure out which type I've gotten from a union type or something reasonably specific, and here I'm skirting around the `Instance` type being defined as `WhisperComponent<WhisperComponentType>`, and in this particular case the `child` is not actually part of that type because I've introduced a psuedo-component that doesn't actually exist in the Olive Helps context.

... though as I type this out, I'm wondering whether I should instead include the Whisper psuedo-component in the `Instance` type and guard against it elsewhere.

I'll play around with it and see what that looks like

ldk/javascript/src/whisper/react/renderer-config.ts
richseviora: Nah you got it right, this was an oversight that I didn't catch. Thanks for the catch :)

ldk/javascript/src/whisper/react/whisper-render-instance.ts
richseviora: Yep! Good catch :)

ldk/javascript/src/whisper/react/whisper-render-instance.ts
richseviora: This was an interesting area and I'm not 100% sure what to do yet. 

When I'm unmounting components after the Whisper has been closed, I'm still getting a final call to this function to update the container with no components.  I don't see a way to "destroy" a container.

ldk/javascript/src/whisper/react/renderer-config.ts
richseviora: I ended up handling this one a bit differently :)

ldk/javascript/src/whisper/react/component-handlers.ts
```diff
@@ -0,0 +1,255 @@
+/* eslint-disable */
```
cshields-oliveai, line 1: Did you intend to keep the eslint ignoring the entire file?

ldk/javascript/src/whisper/react/readme.md
```diff
@@ -0,0 +1,61 @@
...
+
+**Warning** this is all experimental and there may be bugs. Please open issues as you run into them!
+
+You can now use React to manage your Whispers! In just the same way you can use React to create and update complex user interfaces in HTML, you can use React to create and update Whispers in Olive Helps.
+
+## Technical Summary
+
+We've developed a custom React Host Configuration that handles creating and updating Whispers. It treats each Whisper as a separate container.
+
+TODO: Add more words.
```
cshields-oliveai, line 11: leave this TODO in?

ldk/javascript/src/whisper/react/component-types.ts
```diff
@@ -0,0 +1,81 @@
+import { ReactNode } from 'react';
+import * as Whisper from '../types';
+import { WhisperHandler } from '../types';
+
+/**
+ * @internal
```
cshields-oliveai, line 6: Will these annotations prevent these from being picked up in our generated documentation?

ldk/javascript/src/whisper/react/component-types.ts
```diff
@@ -0,0 +1,81 @@
...
+  'oh-select': SelectProps;
+  'oh-section-title': SectionTitleProps;
+  'oh-telephone': TelephoneProps;
+  'oh-text-input': TextInputProps;
+  'oh-whisper': WhisperProps;
+}
+
+declare global {
+  // eslint-disable-next-line @typescript-eslint/no-namespace -- needed to populate jsx components
+  namespace JSX {
```
cshields-oliveai, line 78: Is this merging in the `HelpsComponents` types into the `JSX` namespace? Can you explain this declaration?

ldk/javascript/src/whisper/react/renderer-config.ts
```diff
@@ -0,0 +1,336 @@
...
+  supportsMutation: boolean;
+  supportsPersistence: boolean;
+  supportsHydration: boolean;
+
+  getInstanceFromNode(node: any): any;
+
+  isOpaqueHydratingObject(value: any): boolean;
+
+  makeOpaqueHydratingObject: any;
+  akeOpaqueHydratingObject: any;
```
cshields-oliveai, line 101: Was it intended to include both `makeOpaqueHydratingObject` and `akeOpaqueHydratingObject`? Seems like a refactor gone wrong.

ldk/javascript/src/whisper/react/renderer.test.tsx
```diff
@@ -0,0 +1,357 @@
...
+      reject = reject1;
+    });
+    return {
+      promise,
+      resolve: resolve!,
+      reject: reject!,
+    };
+  }
+
+  describe('writing new whisper', () => {
```
cshields-oliveai, line 76: Good tests.

ldk/javascript/src/whisper/react/renderer.test.tsx
```diff
@@ -0,0 +1,357 @@
...
+      );
+      await deferred.promise;
+
+      const button = mocked(whisperInterface.createOrUpdateWhisper).mock.calls[0][0]
+        .components[0] as Button;
+      button.onClick(undefined, null as any);
+
+      expect(whisperInterface.createOrUpdateWhisper).toHaveBeenCalledWith({
+        label: 'whisper.label',
+        onClose: expect.any(Function),
```
cshields-oliveai, line 265: Could you check that the same `onClose` function is being sent to the whisperInterface? Or is the reference to that function being lost in mapping?

cshields-oliveai: Probably don't have to cover it in every test but it could be worthwhile instead of just `expect.any(Function)`

cshields-oliveai (approved):

ldk/javascript/src/whisper/react/component-handlers.ts
richseviora: Nope! Thanks for the catch!

ldk/javascript/src/whisper/react/component-types.ts
richseviora: When parsing JSX, Typescript looks into the `JSX` namespace in the global scope and accesses the `IntrinsicElements` interface to see what are valid elements.

I initially added the elements in the `HelpsComponent` interface directly, but that meant I couldn't differentiate between the HTML elements added in the `@types/react` package in other contexts, so I pulled it out here.

ldk/javascript/src/whisper/react/renderer-config.ts
richseviora: Nope, it was a typo or something! Gone

ldk/javascript/src/whisper/react/renderer.test.tsx
richseviora: The identity does change here, I intercept it in the WhisperInstanceWrapper

ldk/javascript/src/whisper/react/readme.md
richseviora: Yeah I'll tackle additional documentation in HELPS-1008

ldk/javascript/src/whisper/react/component-handlers.ts
```diff
@@ -0,0 +1,258 @@
...
+    case WhisperComponentType.Select:
+      return selectHandler;
+    case WhisperComponentType.Telephone:
+      return telephoneHandler;
+    case WhisperComponentType.TextInput:
+      return textInputHandler;
+    case WhisperComponentType.SectionTitle:
+      return sectionTitleHandler;
+    default:
+      assertUnreachable(type);
```
RusGraf, line 198: I understand that it is a last statement, but for the sake if anyone will be extending this functionality, will it make more sense to add `break;` at the end of the default case?

ldk/javascript/src/whisper/react/renderer-config.ts
```diff
@@ -0,0 +1,337 @@
...
+  ): Instance;
+
+  cloneHiddenTextInstance(
+    instance: Instance,
+    text: string,
+    internalInstanceHandle: any,
+  ): TextInstance;
+}
+
+export const config: CoreConfig & PersistenceConfig = {
```
RusGraf, line 150: Do we have a sufficient unit test coverage for this? I see some business logic and it worries me that we could fat finger/brake it.

ldk/javascript/src/whisper/react/renderer-config.ts
```diff
@@ -0,0 +1,337 @@
...
+      value.key = internalInstanceHandle.key;
+    }
+    getHandlerByTagType(type).assignTextChildren?.(value, newProps);
+    return value;
+  },
+  createContainerChildSet(container: Container): ChildSet {
+    return {
+      label: '',
+      components: [],
+      // eslint-disable-next-line @typescript-eslint/no-empty-function
```
RusGraf, line 217: this is pretty minor, but documentation recommends to do:
`
onclose: () => {
  // do nothing.
}
`
https://eslint.org/docs/rules/no-empty-function

ldk/javascript/src/whisper/react/renderer-config.ts
richseviora: This configuration is consumed by the renderer which we test thoroughly; if we were to make any breaking changes here (like failing to return created objects) then that'll show up in those tests as whispers that weren't correctly created/updated.

ldk/javascript/src/whisper/types.ts
```diff
@@ -1,3 +1,10 @@
+/**
+ * REMINDER: Whenever you add (or remove) components from this file, you MUST
```
RusGraf, line 2: Nice, thanks for adding this comment.

ldk/javascript/src/whisper/react/renderer-config.ts
RusGraf: I see.

RusGraf (approved):

PR #361, opened 2021-09-24
Helps 1515 add normal to justify content

MaggieMaSS:
justify-content is non-optional but it doesn’t have its value normal added as an option. So normal is added to justify-content!

Please note : Normal is initial value of CSS justify-content style, the items are packed in their default position as if no justify-content value was set( but we have to assign something to justify-content on our code ). Normal behaves as stretch in flex containers. It will stretch vertically if the height of box is larger than the height of that div . I don’t think items are stretchable right now, but we should be careful to use this normal if we are going to support some zooming or stretching features.


ldk/javascript/examples/self-test-loop/src/tests/whisper/index.ts
```diff
@@ -1901,15 +1901,49 @@ export const testAlignItems = (): Promise<boolean> =>
...AAAAAAAABTIgJ45kaZ5oqq5s674we0VS9l4JICybCxUIh4HiMzwagorroiAEeDeJpBerWq/YrPYUAgAh+QQAAwAAACwAAAAAFAAUAIRiGv9jHP9kHf9kHv9oI/9tK/9xMf90NP93Of9+Q/9/Rf+ETP+FTv+KVf+VZf+pgv+0kv+2lv+4mP+6mv+7m/+8nf/k2P/l2f/m2v/n3P/o3v/p3/////8AAAAAAAAAAAAFMiAnjmRpnmiqrmzrvjB7RVL2XgkgLJsLFQjHgeIzPBqCiuuiIAR4N4mkF6tar9is9hQCACH5BAADAAAALAAAAAAUABQAhGIa/2Mc/2Qd/2Qe/2gj/20r/3Ex/3U2/3c5/35D/39F/4RM/4VO/4pU/5Vl/6mC/7SS/7eW/7iY/7qa/7ub/7yd/+TY/+XZ/+ba/+fc/+je/+nf/////wAAAAAAAAAAAAUyICeOZGmeaKqubOu+MHtFUvZeCSAsmwsVCMeh4jM8GgJi66IgBHg3iaQXq1qv2Kz2FAIAIfkEAAMAAAAsAAAAABQAFACEYhr/Yxz/ZB3/ZR7/aCP/bSv/cTH/dTf/dzn/fkP/f0X/hEz/hk7/ilX/lmb/qYL/tJL/t5b/uJj/upr/upv/vJ3/5Nj/5dn/5tr/59z/6N7/6d//////AAAAAAAAAAAABTIgJ45kaZ5oqq5s674we0VS9l4JICybCxUIx6HiMzwaAmLroiAEeDeJpBerWq/YrPYUAgAh+QQAAwAAACwAAAAAFAAUAIRiGv9jHP9kHf9lHv9oI/9tK/9xMf91N/93Of9+Q/9/Rf+ETP+GTv+KVf+WZv+pgv+0kv+3lv+4mP+6mv+6m/+8nf/k2P/l2f/m2v/n3P/o3v/p3/////8AAAAAAAAAAAAFMiAnjmRpnmiqrmzrvjB7RVL2XgkgLJsLFQjHoeIzPBoCYuuiIAR4N4mkF6tar9is9hQCACH5BAADAAAALAAAAAAUABQAhGIa/2Mc/2Qd/2Ue/2gj/20r/3Ex/3U3/3c5/35D/39F/4RM/4ZO/4pV/5Zm/6mC/7SS/7eW/7iY/7qa/7qb/7yd/+TY/+XZ/+ba/+fc/+je/+nf/////wAAAAAAAAAAAAUyICeOZGmeaKqubOu+MHtFUvZeCSAsmwsVCMeh4jM8GgJi66IgBHg3iaQXq1qv2Kz2FAIAOw==) `;
```
RusGraf, line 1901: I know that this was not introduced by this PR, but could you please move it to a file with constants?

ldk/javascript/examples/self-test-loop/src/tests/whisper/index.ts
```diff
@@ -2537,3 +2616,130 @@ export const testScrollInsideBox = (): Promise<boolean> =>
...
```
RusGraf, line 2622: Looks like this is the same image as above, You should be able to reuse it after abstracting it to another file.

ldk/javascript/examples/self-test-loop/src/tests/whisper/index.ts
```diff
@@ -2537,3 +2616,130 @@ export const testScrollInsideBox = (): Promise<boolean> =>
       reject(error);
     }
   });
+
+export const testJustifyContent = (): Promise<boolean> =>
```
RusGraf, line 2620: What is the difference between this test and testAllignItems?tThey both seems to have `alignItems` and `justifyContent` configuration. Do we need both tests?

RusGraf (approved):

epiccoleman (approved):

PR #371, opened 2021-10-04
Expanded padding test to help test out input labels

MichaelOliveMcCarrick:


ldk/javascript/examples/self-test-loop/src/tests/whisper/index.ts
```diff
@@ -2313,6 +2316,27 @@ export const testAutocomplete = (): Promise<boolean> =>
 export const testPadding = (): Promise<boolean> =>
   new Promise(async (resolve, reject) => {
     try {
+      const divider = createDivider();
```
MaggieMaSS, line 2319: divider is a component, you can use that like this       
` {
          type: WhisperComponentType.Divider,
        },`

Might be me not fully understanding the requirement, but why do we want to use this `createDivider()` function to get divider instead of using divider component directly?

MichaelOliveMcCarrick: Just me making it slightly cleaner. No real difference.

MaggieMaSS (approved):

JosefBud-olive (approved):

PR #407, opened 2021-11-04
increment oliveHelpsContractVersion to 0.1.7

MaggieMaSS:
Increment oliveHelpsContractVersion to 0.1.7 for the incoming screen aptitude.


JosefBud-olive (approved):

MaggieMaSS:
UI is  on 0.1.6 and doesn’t have support for screen , if users install those loops, it could crash the app/perform unexpectedly if we don’t bump the version with this new aptitude. After discussion with Sachin and Billy, we decided to increase our oliveHelpsContractVersion to 0.1.7

MichaelOliveMcCarrick (approved):

PR #444, opened 2021-11-24
Helps 2090 progress whisper component

MaggieMaSS:
Add `Progress` component  to Whisper Component


MichaelOliveMcCarrick (approved):

MaggieMaSS:
Screen Shot 2021-11-29 at 5 04 39 PM
>

MaggieMaSS:
Screen Shot 2021-11-29 at 5 04 39 PM

MaggieMaSS:
https://user-images.githubusercontent.com/84045424/143951652-ff38e43f-f3af-4b4e-b9b2-df40433cf99a.mov

ldk/javascript/src/types/oliveHelps/whisper.d.ts
```diff
@@ -54,6 +55,8 @@ declare namespace WhisperService {
 
   type IconSize = 'small' | 'medium' | 'large' | 'x-large';
 
+  type Shape = 'circular' | 'linear';
```
BrettMugglinOlive, line 58: `Shape` is a pretty broad type, and one that we probably want to make more specific to the Progress component, just in case. So maybe it should be `ProgressShape` or something like that.

ldk/javascript/src/whisper/types.ts
```diff
@@ -46,6 +46,10 @@ export enum WhisperComponentType {
    * The password input field allows the user to provide a password. This field protects the user by obscuring what they type. Showing each character as a solid black dot.
    */
   Password = 'password',
+  /**
+   * The progress field allows the user to see the visualized progress indicator.
```
BrettMugglinOlive, line 50: For this description, maybe we should take some inspiration from the [material ui component ](https://mui.com/components/progress/) it's based off of.

It could read "Progress components express an unspecified wait time or display the length of a process."

ldk/javascript/examples/self-test-loop/src/tests/whisper/index.ts
```diff
@@ -2896,3 +2894,41 @@ export const testMissingLayouts = (): Promise<boolean> =>
...
+  new Promise(async (resolve, reject) => {
+    await whisper.create({
+      label: 'Did progress indicator have indefinite animation? ',
+      onClose: () => {
+        console.debug('closed');
+      },
+      components: [
+        {
+          type: WhisperComponentType.Progress,
+          size: StyleSize.Medium,
```
BrettMugglinOlive, line 2925: For the purposes of this test, could you try this out with all the different sizes and shapes?

ldk/javascript/examples/self-test-loop/src/tests/whisper/index.ts
```diff
@@ -2896,3 +2894,41 @@ export const testMissingLayouts = (): Promise<boolean> =>
...
+  new Promise(async (resolve, reject) => {
+    await whisper.create({
+      label: 'Did progress indicator show the percentage? ',
+      onClose: () => {
+        console.debug('closed');
+      },
+      components: [
+        {
+          type: WhisperComponentType.Progress,
+          determinate: 42,
```
BrettMugglinOlive, line 2908: Could this be done with both different shape types?

BrettMugglinOlive (changes requested):

MaggieMaSS:
Screen Shot 2021-12-01 at 10 52 00 AM

ldk/javascript/src/whisper/types.ts
```diff
@@ -444,6 +452,12 @@ export type RichTextEditor = WhisperComponent<WhisperComponentType.RichTextEdito
 
 export type Divider = WhisperComponent<WhisperComponentType.Divider>;
 
+export type Progress = WhisperComponent<WhisperComponentType.Progress> & {
+  determinate?: number;
+  shape?: ProgressShape;
+  size?: StyleSize;
```
BrettMugglinOlive, line 458: My last question is about the choice to use `StyleSize` here. `none` is an option for that type, so what happens if someone uses that as a value?

BrettMugglinOlive (approved):

PR #457, opened 2021-12-04
[HOLD] Helps-2438 Add filterOptions parameter to the Autocomplete component

BrettMugglinOlive:


ldk/javascript/src/types/oliveHelps/whisper.d.ts
```diff
@@ -79,7 +79,7 @@ declare namespace WhisperService {
 
   interface Whisper {
     id: string;
-    close: Common.Readable<undefined>;
+    close: (cb: (err: Error | undefined) => void) => void;
```
MichaelOliveMcCarrick, line 82: Why did this get changed?

ldk/javascript/src/whisper/mapper.ts
```diff
@@ -106,6 +106,7 @@ export function mapToInternalChildComponent(
       // eslint-disable-next-line
       const { onChange, onSelect, options } = component;
       const dataObj = { data: options };
+
```
MichaelOliveMcCarrick, line 109: Extra new line

MichaelOliveMcCarrick (approved):

MaggieMaSS (commented):
https://crosschx.atlassian.net/browse/HELPS-2447 
Seems like the Sensors&Aptitude (goja layer) ticket is not taken 🤔 And we need to wait for their implementation.

ldk/javascript/src/types/oliveHelps/whisper.d.ts
```diff
@@ -113,7 +113,17 @@ declare namespace WhisperService {
     value: string;
   };
 
+  interface AutocompleteFilterOptions {
+    ignoreAccents?: boolean;
+    ignoreCase?: boolean;
+    limit?: number;
+    matchFrom?: 'any' | 'start';
+    stringify?: (option: AutocompleteOption) => void;
```
JosefBud-olive, line 121: Based on the component's code, I think stringify should be required here, or made optional in the UI

```js
  const parsedFilterOptions = filterOptions ? createFilterOptions({
    ...filterOptions,
    stringify: option => filterOptions.stringify(option),
  }) : undefined;
```

ldk/javascript/src/types/oliveHelps/whisper.d.ts
JosefBud-olive: Also wondering this? Is this possibly leftover from when you were testing different ways to type the stringify function?

JosefBud-olive (changes requested):
`stringify` not being required on our end, or leaving it required in the UI, will definitely cause some issues for folks so we should figure out what we want to do there

PR #459, opened 2021-12-08
[HELPS-2404] Fetch polyfill example.

MichaelOliveMcCarrick:
This is an example as to how you can polyfill fetch with the ldk. This may be useful if you use extra libraries that depend on WHATWG's fetch Api.

### DIFF
--- a/ldk/javascript/examples/fetch/package.json
+++ b/ldk/javascript/examples/fetch/package.json
@@ -1,0 +1,32 @@
+{
+  "name": "fetch-polyfill-example",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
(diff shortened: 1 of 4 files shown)


ldk/javascript/examples/fetch/package.json
```diff
@@ -0,0 +1,32 @@
...
+          {
+            "value": "httpstat.us"
+          }
+        ]
+      },
+      "whisper": {}
+    }
+  },
+  "dependencies": {
+    "@oliveai/ldk": "file:../..",
```
BrettMugglinOlive, line 25: Should this be a specific version? Or should we keep it as the relative path?

BrettMugglinOlive (approved):

PR #460, opened 2021-12-08
updated ocr test

MaggieMaSS:
If we filter out lower conv results, there's no value returned. So I deleted the filter function so that QA can see some results there first. Will figure out what happened there.

### DIFF
--- a/ldk/javascript/examples/self-test-loop/src/tests/screen/index.ts
+++ b/ldk/javascript/examples/self-test-loop/src/tests/screen/index.ts
@@ -1,17 +1,6 @@
 /* eslint-disable no-async-promise-executor */
 import { screen, whisper, window } from '@oliveai/ldk';
 
-export const testOCR = (): Promise<boolean> =>
-  new Promise(async (resolve, reject) => {
-    try {
-      await OcrLoop();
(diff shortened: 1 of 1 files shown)


ldk/javascript/examples/self-test-loop/src/tests/screen/index.ts
```diff
@@ -97,11 +96,10 @@ export async function performOcr() {
...
         console.log('OCR Results: ');
         console.log(JSON.stringify(result));
-
-        console.log(rebuild_image(result));
-        const resFilter = result.filter((res) => res.conf > 75);
-        const resFilterResult = resFilter.map((res) => `${res.text}`);
-        writeWhisper(`result`, `${resFilterResult.join(' ')}`);
+        const concatResult = result.map((res) => res.text).join(' ');
+        console.log('concatResult', concatResult);
+        console.log('result: ', rebuildImage(result));
```
MichaelOliveMcCarrick, line 101: As long as this eventually terminates, looks good to me

MichaelOliveMcCarrick (approved):

JosefBud-olive (approved):

PR #463, opened 2021-12-10
Bread crumbs fix for box components updates

alexmatthew-olive:
Breadcrumbs Links onClick handlers were throwing an error when using the whisper instance from the handler to update the current whisper if the components array had a box component. This treats the links array as a "child component" so the recursive call gets called for each link


ldk/javascript/src/whisper/mapper.ts
```diff
@@ -4,6 +4,7 @@ import {
   Component,
   NewWhisper,
   StateMap,
+  Breadcrumbs,
```
MaggieMaSS, line 7: Correct me if I'm wrong, but I think we don't need to import Breadcrumbs, since we can use WhisperComponentType.Breadcrumbs below.  0.0

ldk/javascript/src/whisper/mapper.ts
```diff
@@ -49,8 +50,13 @@ export function mapToInternalChildComponent(
           : undefined,
       } as WhisperService.Box;
     case WhisperComponentType.Breadcrumbs:
+      // eslint-disable-next-line no-case-declarations
+      const { links, ...props } = component as Breadcrumbs;
```
MaggieMaSS, line 54: `const { links, ...props } = component as WhisperComponentType.Breadcrumbs` Maybe this is better? So we don't need to import Breadcrumb. 0.0

ldk/javascript/src/whisper/mapper.ts
alexmatthew-olive: When I tried that I got this error `Conversion of type 'Breadcrumbs' to type 'WhisperComponentType.Breadcrumbs' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first` and I was following what Box did on line 37

ldk/javascript/src/whisper/mapper.ts
alexmatthew-olive: Responded on other comment

MaggieMaSS (approved):

ldk/javascript/src/whisper/mapper.ts
MaggieMaSS: I realized that you adjust Breadcrumbs's order a little bit, :) they are  alphabetical ordered

BrettMugglinOlive (approved):

PR #472, opened 2021-12-21
Create screen test with ocr

MaggieMaSS:
Added more selftest using cursor aptitude


ldk/javascript/examples/self-test-loop/src/config/screen-group.ts
```diff
@@ -35,4 +35,10 @@ export const screenTestGroup = (): TestGroup =>
       10000,
       'Please open a new fullscreen window and change the contents (A browser works well for this).',
     ),
+    new LoopTest(
+      'Screem Aptitude - OCRUsingCursor ',
```
MichaelOliveMcCarrick, line 39: Screem -> Screen

ldk/javascript/examples/self-test-loop/src/tests/screen/index.ts
```diff
@@ -109,3 +111,101 @@ export async function performOcr() {
       });
   });
 }
+
+console.log(`Starting app`);
```
MichaelOliveMcCarrick, line 115: Random console log

ldk/javascript/examples/self-test-loop/src/tests/screen/index.ts
```diff
@@ -109,3 +111,101 @@ export async function performOcr() {
...
+        }
+
+        i += 1;
+
+        if (i >= 2) {
+          cursorPositionStream.cancel();
+        }
+      }
+    })
+    .then((cancellable: Cancellable) => {
```
MichaelOliveMcCarrick, line 148: Since we're already doing async/await, can we await for the cancellable instead of promise chaining?

ldk/javascript/examples/self-test-loop/src/tests/screen/index.ts
```diff
@@ -109,3 +111,101 @@ export async function performOcr() {
...
+async function testOCRUsingCursor() {
+  let i = 0;
+  let cursorPositionStream: Cancellable;
+  let topParam: number;
+  let leftParam: number;
+  let topParam1: number;
+  let leftParam1: number;
+  await sleep(1000);
+  cursor
+    .listenPosition((response) => {
```
MichaelOliveMcCarrick, line 126: Looks like listenPosition simply polls the cursor position every .5 seconds, that seems like it might be difficult to setup a box around specific text.

ldk/javascript/examples/self-test-loop/src/tests/screen/index.ts
```diff
@@ -109,3 +111,101 @@ export async function performOcr() {
...
+    top: topParam,
+    left: leftParam,
+    width: parseInt(width, 10),
+    height: parseInt(height, 10),
+  };
+  console.log('got OCR coordinates:');
+  console.log(ocrCoordinates.top, ocrCoordinates.left, ocrCoordinates.width, ocrCoordinates.height);
+  console.log('performing ocr with coordinates...');
+
+  screen
```
MichaelOliveMcCarrick, line 165: Same as comment above about async/await

ldk/javascript/examples/self-test-loop/src/tests/screen/index.ts
```diff
@@ -109,3 +111,97 @@ export async function performOcr() {
...
+        console.debug(` topParam - Y - ${topParam1}`);
+      }
+      if (i === 1) {
+        topParam1 = response.y;
+        leftParam1 = response.x;
+        console.debug(`leftParam1 - X - ${leftParam1}`);
+        console.debug(` topParam1 - Y - ${topParam1}`);
+      }
+
+      i += 1;
```
MichaelOliveMcCarrick, line 137: This is really my only issue with the PR, listenPosition has a .5 second polling time. I'd rather do something like have the user hit spacebar to capture the cursor position, then hit spacebar again to capture the second point instead of polling twice on the listener.

MaggieMaSS: @MichaelOliveMcCarrick  Sry for the late reply, yeah, I think that's a better way to initiate the process!

JosefBud-olive:
Looks like the package-lock was massively changed in this PR, can you undo the changes to that? You may need to use node v16, it looks like you may be on an earlier version and it's changing the file.

PR #474, opened 2021-12-29
HELPS-2776 onClose is not called for react whispers

### LINKED ISSUE #473: onClose is not called for react whispers
LDK: Version 3.9.0

The onClose handler is not being called for react whispers. Here is a very simple test that I set up:

```
    await ReactWhisper.renderNewWhisper(
      <oh-whisper label="test" onClose={() => console.log('closing whisper')}>
        <oh-markdown body="test" />
      </oh-whisper>,
    );
```

When I close the whisper, here is the output in logs:
```
021-12-29 10:40:37.872] [ui] [debug] sending tracking event
[2021-12-29 10:40:37.883] [ui] [debug] sending tracking event
[2021-12-29 10:40:37.000] [desktop-core] [debug] whisper closed interactionType=close whisperType=universal
[2021-12-29 10:40:58.000] [desktop-core] [debug] Sidenote core loop listener called
[2021-12-29 10:40:58.000] [desktop-core] [debug] matched 0 notes for keyword
[2021-12-29 10:40:58.000] [desktop-core] [debug] on-boarding loop clipboard listener called
[2021-12-29 10:40:58.475] [ui] [debug] sending tracking event
```

Using the same test without react, my log statement is just after
`[2021-12-29 10:40:37.000] [desktop-core] [debug] whisper closed interactionType=close whisperType=universal`

ssteuteville:
Fixes #473

Added a way for whisper render instances to track and call an internal close handler (used for unmounting stuff) and an external on close handler (for the end user to use).

If you check my previous commit it used a different approach, I think this one is better though.

I also question that `render` and `renderNewWhisper` should take a ReactNode, I think `ReactElement` would be more accurate since the top level component does need to be a `ReactElement`.

### DIFF
--- a/ldk/javascript/package-lock.json
+++ b/ldk/javascript/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@oliveai/ldk",
-  "version": "3.7.0",
+  "version": "3.9.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@oliveai/ldk",
-      "version": "3.7.0",
+      "version": "3.9.0",
       "license": "MIT",
       "dependencies": {
         "@babel/core": "^7.15.8",
--- a/ldk/javascript/src/whisper/react/renderer.ts
+++ b/ldk/javascript/src/whisper/react/renderer.ts
@@ -1,5 +1,5 @@
 import * as Reconciler from 'react-reconciler';
-import { ReactElement } from 'react';
+import { ReactNode } from 'react';
 import { config } from './renderer-config';
 import { WhisperRenderingInterface, WhisperRenderInstance } from './whisper-render-instance';
 // This side effect import makes the JSX namespace addition available without
@@ -13,13 +13,9 @@ import { WhisperInstance, WhisperInstanceWrapper } from './whisper-instance-wrap
 const renderer = Reconciler.default(config);
 
 export async function render(
-  element: ReactElement,
+  element: ReactNode,
   whisperInterface: WhisperRenderingInterface,
 ): Promise<WhisperInstance> {
-  if (typeof element.props.onClose === 'function') {
-    whisperInterface.setOnClose(element.props.onClose)
-  }
-
   // Tag here drives what sort of "modes" its using. 0 = LegacyRoot.
   const container = renderer.createContainer(whisperInterface, 0, false, null);
   const wrapper = new WhisperInstanceWrapper(container, whisperInterface, renderer);
@@ -31,7 +27,7 @@ export async function render(
  *
  * @param element
  */
-export function renderNewWhisper(element: ReactElement): Promise<WhisperInstance> {
+export function renderNewWhisper(element: ReactNode): Promise<WhisperInstance> {
   return render(element, new WhisperRenderInstance());
 }
 
--- a/ldk/javascript/src/whisper/react/renderer.ts
+++ b/ldk/javascript/src/whisper/react/renderer.ts
@@ -1,5 +1,5 @@
 import * as Reconciler from 'react-reconciler';
-import { ReactNode } from 'react';
+import { ReactElement } from 'react';
 import { config } from './renderer-config';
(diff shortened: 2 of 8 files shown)


ssteuteville:
there are a bunch of whitespace changes I wasn't expecting, which is weird since I ran `npm run lint`.. I'll get those fixed!

ldk/javascript/package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "@oliveai/ldk",
-  "version": "3.7.0",
+  "version": "3.9.0",
```
ssteuteville, line 3: not sure about these changes, npm install did it

ldk/javascript/src/whisper/react/renderer.ts
```diff
@@ -13,9 +13,13 @@ import { WhisperInstance, WhisperInstanceWrapper } from './whisper-instance-wrap
 const renderer = Reconciler.default(config);
 
 export async function render(
-  element: ReactNode,
+  element: ReactElement,
```
ssteuteville, line 16: see description for explanation of this type change

ssteuteville: ignore this, I changed approach

ldk/javascript/package-lock.json
iangallOlive: I think this is fine. It looks like someone updated the `package.json` version without updating `package-lock.json`'s (or running `npm i`). So while it's not related, I think this is good to include

iangallOlive:
I don't have enough experience with React whispers, but this *looks* good. And I appreciate you including a test to prove it's working now, so thanks for doing that!

ssteuteville:
> I don't have enough experience with React whispers, but this _looks_ good. And I appreciate you including a test to prove it's working now, so thanks for doing that!

Thanks @iangallOlive let me know if there is anything else I can do or anyone I can poke to get this merged and included in a release please!

iangallOlive:
@ssteuteville Two quick things.

I enabled the build workflow for this PR, so those are running now. One failed, `prettier-check`, so you'll need to check that out and fix whatever thing (small, I'm sure) it's complaining about.

And then - I don't have experience with JSX whispers, but I **think** the `onClose` property is required (even if it's an "empty" callback). Could you verify compilation fails if you don't provide that property. And if it's _not_ required, add a test to make sure your changes still work if you _don't_ provide an `onClose` function. I'm not exactly sure what the test would do, but still. If it is required, ignore this request :)

Either way, we're working on getting reviews and figuring out a release plan for this.

ssteuteville:
@iangallOlive I ran prettier! looks like the workflows need to be approved again. It would be awesome if someone had time to write a quick developer guide, i was running lint in the javascript module rather than at top level!

here is what i get when i don't pass on close to the whisper:
```
[tsl] ERROR in /Users/ssteuteville/dev/sonar-loop/src/v2/common/components/viewManager/ViewManagerWhisper.tsx(15,8)
      TS2741: Property 'onClose' is missing in type '{ children: Element; label: string; }' but required in type 'WhisperProps'.
 @ ./src/v2/common/components/viewManager/hooks/useCreateViewManager.ts 25:43-75
 @ ./src/v2/common/components/viewManager/hooks/index.ts 14:29-62
 @ ./src/v2/common/components/viewManager/index.ts 26:13-31
 @ ./src/v2/common/components/index.ts 51:13-37
 @ ./src/v2/loops/circulator/whisper.tsx 21:19-53
 @ ./src/v2/loops/circulator/index.tsx 75:32-52
 @ ./src/services/instancePool.ts 19:35-68
 @ ./src/index.ts 34:37-71
```

So, I think we're good on that front! Thanks for being attentive here :)

iangallOlive:
@ssteuteville Sorry, just approved/enabled the workflows again!
I will get a card created for us to add dev info to the repo.
And thanks for confirming `onClose` is required and fails to compile without!
I'll message you on Slack about getting this released

BrettMugglinOlive (approved):

iangallOlive (approved):

PR #475, opened 2022-01-03
HELPS-2529-  Update "User Aptitude - JWT - Email Claim" test package.json

MaggieMaSS:
When running the "User Aptitude - JWT - Email Claim" test, the test initiates, but then freezes and never completes.
It turns out that we need to add the `optionalClaims` and add the email address to it.
The permission for user in `package.json` should be like this: 
```

      "user": {
        "optionalClaims": [
          {          
            "value": "email"
          }
        ]
      },
```

### DIFF
--- a/ldk/javascript/examples/self-test-loop/package.json
+++ b/ldk/javascript/examples/self-test-loop/package.json
@@ -47,7 +47,7 @@
       "user": {
         "optionalClaims":[
           {
-            "value" :"team-ldx@oliveai.com"
+            "value" :"email"
           }
         ]
       },
--- a/ldk/javascript/examples/self-test-loop/package.json
+++ b/ldk/javascript/examples/self-test-loop/package.json
@@ -47,10 +47,9 @@
       "user": {
         "optionalClaims":[
           {
-            "value" :"add your email address here"
+            "value" :"team-ldx@oliveai.com"
           }
         ]
-        
       },
       "system": {},
       "screen": {},
--- a/ldk/javascript/examples/self-test-loop/package.json
+++ b/ldk/javascript/examples/self-test-loop/package.json
@@ -44,7 +44,14 @@
       "filesystem": {},
       "process": {},
       "ui": {},
-      "user": {},
+      "user": {
+        "optionalClaims":[
+          {
+            "value" :"add your email address here"
+          }
+        ]
+        
+      },
       "system": {},
       "screen": {},
       "keyboard": {},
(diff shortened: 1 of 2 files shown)


ldk/javascript/examples/self-test-loop/package.json
```diff
@@ -44,7 +44,14 @@
...
       "process": {},
       "ui": {},
-      "user": {},
+      "user": {
+        "optionalClaims":[
+          {
+            "value" :"add your email address here"
+          }
+        ]
+
```
JosefBud-olive, line 53: Extra line break here

ldk/javascript/examples/self-test-loop/package.json
```diff
@@ -44,7 +44,14 @@
       "filesystem": {},
       "process": {},
       "ui": {},
-      "user": {},
+      "user": {
+        "optionalClaims":[
+          {
+            "value" :"add your email address here"
```
JosefBud-olive, line 50: Should this include an actual email address for the self test loop?

JosefBud-olive (commented):
Could you add a description? 😁

ldk/javascript/examples/self-test-loop/package.json
MaggieMaSS: I used my email on my end 0.0. Do you volunteer to put your email here? Or I can just put Brett's 0,0

JosefBud-olive: @MaggieMaSS if we use a Google Group email for our team, that should do the trick!

ldk/javascript/examples/self-test-loop/src/config/user-group.ts
```diff
@@ -21,6 +21,6 @@ export const userTestGroup = (): TestGroup =>
       'User Aptitude - JWT - Email Claim',
       userTests.testJwtIncludeEmail,
       10000,
-      'No action required',
+      'Please check package.json and update the value of email to your email address!',
```
JosefBud-olive, line 24: This is outdated now that we added an e-mail address

ldk/javascript/examples/self-test-loop/package.json
```diff
@@ -44,7 +44,13 @@
       "filesystem": {},
       "process": {},
       "ui": {},
-      "user": {},
+      "user": {
+        "optionalClaims":[
+          {
+            "value" :"team-ldx@oliveai.com"
```
MichaelOliveMcCarrick, line 50: Shouldn't this be 'email'? Does this actually work?

MichaelOliveMcCarrick (approved):

JosefBud-olive (approved):
Look like the space is in the wrong place (`value :email` instead of `value: email`) but not going to hold up the PR over it

PR #479, opened 2022-01-05
HELPS-1992: Allow developers to be able to only install the LDK to use JSX whispers

JosefBud-olive:
[HELPS-1992](https://crosschx.atlassian.net/browse/HELPS-1992)

* Added all of the JSX whisper dependencies to LDK so that they're automatically installed with the LDK, including:
  * `react-dom`
  * `@types/react`
  * `@types/react-reconciler`
  * `@babel/plugin-transform-react-jsx`
  * `react` and `react-reconciler` were already included
* Added `react` alias to the default webpack config
* Added `React` to our root exports so that developers can use our specific version of React, e.g. `import { React } from '@oliveai/ldk'`
* Added `ReactWhisper` to our root exports so that developers can more easily import it, e.g. `import { ReactWhisper } from '@oliveai/ldk'` instead of `import * as ReactWhisper from '@oliveai/ldk/dist/whisper/react/renderer'`
* Updated `husky` and added pre-commit hooks to run `lint-staged`
* Removed the Go LDK GitHub Actions
* Removed old files/folders from `.gitignore`


MichaelOliveMcCarrick (approved):

MaggieMaSS (approved):

iangallOlive:
Everything looks good (and I'm loving the cleanup). I don't think I really understood what loops currently have to do for JSX whispers in regards to React dependencies, but I'm confused about what loops should be doing _after this PR_. Should they not add React dependencies to their loop? Just wondering how loops will/should change

JosefBud-olive:
> Everything looks good (and I'm loving the cleanup). I don't think I really understood what loops currently have to do for JSX whispers in regards to React dependencies, but I'm confused about what loops should be doing _after this PR_. Should they not add React dependencies to their loop? Just wondering how loops will/should change

@iangallOlive This should remove all of the beginning steps of adding react-related dependencies to the loop and its webpack config. It basically gets rid of half the setup shown here: https://docs.oliveai.dev/aptitudes/whisper/jsx-whispers. Besides being easier on devs during setup, this also has the benefit of not having to troubleshoot why someone's JSX whispers aren't working if the dependencies come from us.

iangallOlive:
@JosefBud-olive Gotcha, that sounds good. I guess what I want to make sure of, is that there isn't some breaking change for existing loops if all they did was update their LDK dependency to the version with this PR's changes. I don't **expect** there to be a breaking change, but I wanted to double-check.

And then I feel like we should document somewhere (whether that be the release notes, the dev docs, whatever) to say something like "If you followed previous setup for using JSX whispers, here's what you can undo/change".

JosefBud-olive:
@iangallOlive Nope! No breaking changes in this, just additional features. Right now people add their own React dependencies, all we're doing here is covering it for them now. But if you already have a JSX loop it'll still continue to work, we haven't blocked that.

Totally agree on updating documentation and adding an extra tidbit in the release notes.

PR #484, opened 2022-01-06
Helps 1803 rte value

MichaelOliveMcCarrick:
Temp PR for figuring things out


ldk/javascript/examples/self-test-loop/src/tests/whisper/whisper-update.tsx
```diff
@@ -778,3 +778,67 @@ export const testProcessComponent = (): Promise<boolean> =>
       reject(e);
     }
   });
+
+const RichTextEditorUpdates: React.FunctionComponent<TestComponentProps> = (props) => {
```
MichaelOliveMcCarrick, line 782: This is the test thats failling

PR #491, opened 2022-01-19
HELPS-3058 Change stringify to be an array of strings

BrettMugglinOlive:
- Fixes a type error with `filterOptions.stringify` to make it inline with OH.
- Adds a test to self test loop to excercise `filteerOptions`
- Modifies FilterOption.value to be `any` rather than `string`

### DIFF
--- a/ldk/javascript/examples/self-test-loop/src/tests/whisper/index.ts
+++ b/ldk/javascript/examples/self-test-loop/src/tests/whisper/index.ts
@@ -483,7 +483,6 @@ export const testClickableButton = (): Promise<boolean> =>
               label: `Don't click me`,
               onClick: () => console.debug(`Why'd you do that?`),
               type: WhisperComponentType.Button,
-              // size: ButtonSize.Large,
             },
             {
               buttonStyle: ButtonStyle.Text,
@@ -1842,7 +1841,6 @@ export const testSectionTitle = (): Promise<boolean> =>
           {
             body: 'section Title in center',
             type: WhisperComponentType.SectionTitle,
-            // textAlign: TextAlign.Center,
           },
           {
             body: 'section Title on the left',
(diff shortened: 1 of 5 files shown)


ldk/javascript/examples/self-test-loop/src/tests/whisper/index.ts
```diff
@@ -482,7 +483,7 @@ export const testClickableButton = (): Promise<boolean> =>
               label: `Don't click me`,
               onClick: () => console.debug(`Why'd you do that?`),
               type: WhisperComponentType.Button,
-              size: ButtonSize.Large,
+              // size: ButtonSize.Large,
```
JosefBud-olive, line 486: Does this need to be commented out? If we don't need it we should probably just delete it

BrettMugglinOlive: That's probably from me just messing around while writing documentation, and forgetting to clear things. I'll clean that up.

iangallOlive (approved):

ldk/javascript/src/whisper/types.ts
```diff
@@ -202,7 +202,7 @@ export enum IconSize {
 
 export type AutocompleteOption = {
   label: string;
-  value: string;
+  value: any;
```
JosefBud-olive, line 205: Not something to hold up the PR but we should add tests to make sure this actually works with anything other than strings, as far as I can tell we're not testing it?

JosefBud-olive (approved):

ldk/javascript/src/whisper/types.ts
JosefBud-olive: Looks like the build is failing at the lint stage because of this too, might want to disable eslint for this line for now?

BrettMugglinOlive: I did that and it's still failing :/

PR #495, opened 2022-01-20
Add Icons to Button component

MaggieMaSS:
1. Add startIcon and endIcon to Button component. They are optional props.
2.  startIcon appears at the start of button's label, endIcon appears at the end of button's label.
3.  IconVariant is also supported with startIcon and endIcon

Screen Shot 2022-01-20 at 12 52 05 PM


JosefBud-olive:
Looks like there's conflicts after my ticket was merged in, I think there's some overlap in our `IconVariant` enums :)

MichaelOliveMcCarrick (approved):

PR #506, opened 2022-02-01
HELPS-3108 Add navigationType to NavigationEvents

BrettMugglinOlive:
Should have the value "real" if it is a full page load, or "history" if it is an SPA navigation

### DIFF
--- a/ldk/javascript/src/browser/index.ts
+++ b/ldk/javascript/src/browser/index.ts
@@ -5,11 +5,12 @@ import { Cancellable } from '../cancellable';
 import { promisifyListenable, promisifyWithParam } from '../promisify';
 
 export interface NavigationDetails {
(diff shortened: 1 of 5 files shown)


ldk/javascript/src/browser/index.ts
```diff
@@ -5,11 +5,12 @@ import { Cancellable } from '../cancellable';
 import { promisifyListenable, promisifyWithParam } from '../promisify';
 
 export interface NavigationDetails {
-  tabId: number;
-  url: string;
   frameId: number;
+  navigationType: string;
```
MaggieMaSS, line 9: Do we want to create a `type` for navigationType?

MaggieMaSS (approved):
Pre-approve with that `type` question.

PR #523, opened 2022-02-28
Helps 2966 grid component

MaggieMaSS:
Add Grid component & its test. 
Grid component provides ability to add layout grid to whispers. 
Data Contract: 
    alignContent?: AlignContent;
    alignItems?: AlignItems;
    children: Array<ChildComponents>;
    direction?: GridDirection;
    justifyContent?: JustifyContent;
    spacing?: number;
    wrap?: Wrap;
    xs?: string;
  } & (
      | {
          container: boolean;
        }
      | {
          item: boolean;
        }
    );
Screen Shot 2022-02-28 at 10 01 52 AM

### DIFF
--- a/ldk/javascript/examples/self-test-loop/package.json
+++ b/ldk/javascript/examples/self-test-loop/package.json
@@ -11,6 +11,69 @@
   "keywords": [],
   "author": "",
   "ldk": {
+    "configSchema": {
+      "$schema": "http://json-schema.org/draft-07/schema#",
+      "type": "object",
+      "properties": {
+        "testString": {
+          "type": "string",
+          "default": "testing a string"
+        },
+        "testNumber": {
+          "type": "number",
+          "default": 123
+        },
+        "testBoolean": {
+          "type": "boolean",
+          "default": true
+        },
+        "testStringEnum": {
+          "enum": [
+            "this",
+            "is",
+            "sparta"
+          ]
+        },
+        "testNumberEnum": {
+          "enum": [
+            1,
+            2,
+            3
+          ]
+        },
+        "testStringArray": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "uniqueItems": true,
+          "default": []
+        },
+        "testNumberArray": {
+          "type": "array",
+          "items": {
+            "type": "number"
+          },
+          "uniqueItems": true,
+          "default": []
+        },
+        "testObject": {
+          "type": "object",
+          "properties": {
+            "foo": {
+              "type": "string",
+              "default": "testing an object"
+            },
+            "bar": {
+              "type": "string"
+            }
+          }
+        }
+      },
+      "required": [
+        "testVariable"
+      ]
+    },
     "permissions": {
       "browser": {
         "urlDomains": [
--- a/ldk/javascript/examples/self-test-loop/src/config/config-group.ts
+++ b/ldk/javascript/examples/self-test-loop/src/config/config-group.ts
@@ -1,0 +1,14 @@
+import TestGroup from '../testingFixtures/testGroup';
+import { LoopTest } from '../testingFixtures/loopTest';
+
+import * as configTests from '../tests/config';
+
+export const configTestGroup = (): TestGroup =>
+  new TestGroup('Config Aptitude', [
+    new LoopTest(
+      'Config Aptitude - Loop Config Test',
+      configTests.testLoopConfig,
+      10000,
+      'Getting Loop config...',
+    ),
+  ]);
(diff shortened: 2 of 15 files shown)


ldk/javascript/examples/self-test-loop/src/tests/whisper/autocomplete/index.tsx
```diff
@@ -73,6 +73,7 @@ export const testAutocomplete = (): Promise<boolean> =>
             multiple: true,
             onSelect: (error, value, onSelectWhisper) => {
               console.log(`Received selected value: ${JSON.stringify(value)}`);
+              console.log('type of value: ', typeof value);
```
JosefBud-olive, line 76: Is this necessary for testing? The value should always be the same type

ldk/javascript/src/types/oliveHelps/whisper.d.ts
```diff
@@ -187,6 +199,19 @@ declare namespace WhisperService {
...
+    alignItems: AlignItems;
+    children: Array<ChildComponents>;
+    container: boolean;
+    direction?: GridDirection;
+    justifyContent?: JustifyContent;
+    item: boolean;
+    spacing: number;
+    wrap: Wrap;
+    xs?: string;
+  };
```
JosefBud-olive, line 213: I'm not sure if `container` and `item` should be required, in MUI they're mutually exclusive and technically optional. If this is required by the UI we should probably request that be changed.

I don't think `alignItems`, `spacing`, and `wrap` shouldn't be required either.

ldk/javascript/examples/self-test-loop/src/tests/whisper/autocomplete/index.tsx
MaggieMaSS: @JosefBud-olive  Well I believe this is to test a GH issue, and I forgot to delete it 0.0.

ldk/javascript/src/types/oliveHelps/whisper.d.ts
MaggieMaSS: @JosefBud-olive  `container` and `item` are required from UI. But there's logic on UI to deal with this relationship, so I think it's fine that they are required here? 🤔

JosefBud-olive: There's also logic to deal with it if they don't provide either one. In my opinion it would be better a developer experience to write syntax like

```jsx
<oh-grid container>
  <oh-grid item>
  <oh-grid item>
</oh-grid>
```

Rather than

```jsx
<oh-grid container item={false}>
  <oh-grid container={false} item>
  </oh-grid>
</oh-grid>
```

Since this component is supposed to mostly match the MUI Grid component, it would also be more in line with that.

ldk/javascript/src/whisper/types.ts
```diff
@@ -611,6 +636,19 @@ export type Progress = WhisperComponent<WhisperComponentType.Progress> & {
...
+  alignItems?: AlignItems;
+  children: Array<ChildComponents>;
+  container: boolean;
+  direction?: GridDirection;
+  justifyContent?: JustifyContent;
+  item: boolean;
+  spacing?: number;
+  wrap?: Wrap;
+  xs?: string;
+};
```
JosefBud-olive, line 650: This would work for making `container` OR `item` required, making them mutually exclusive

```suggestion
export type Grid = WhisperComponent<WhisperComponentType.Grid> & {
  alignContent?: AlignContent;
  alignItems?: AlignItems;
  children: Array<ChildComponents>;
  direction?: GridDirection;
  justifyContent?: JustifyContent;
  spacing?: number;
  wrap?: Wrap;
  xs?: string;
} & (
  | {
      container: boolean
    }
  | {
      item: boolean
    }
);
```

JosefBud-olive (approved):

ldk/javascript/src/whisper/types.ts
```diff
@@ -122,6 +126,15 @@ export enum AlignItems {
  */
 export const Alignment = JustifyContent;
 
+export enum AlignContent {
```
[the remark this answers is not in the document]
MaggieMaSS: @BrettMugglinOlive  I just found this : https://stackoverflow.com/questions/27539262/whats-the-difference-between-align-content-and-align-items. Hmmm so `alignItem` align along the cross axis within a flex box, but `alignContent `is for multiline within a `flexbox`. So they're similar two different things. 🤔   I think we should set them up separately.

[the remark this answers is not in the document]
MaggieMaSS: @BrettMugglinOlive  I just found this : https://stackoverflow.com/questions/27539262/whats-the-difference-between-align-content-and-align-items. Hmmm so `alignItem` align along the cross axis within a flex box, but `alignContent `is for multiline within a `flexbox`. So they're similar two different things. 🤔   I think we should set them up separately.

BrettMugglinOlive (approved):

PR #530, opened 2022-03-10
HELPS-3478: Loop Config - Add Config Aptitude (LDX)

JosefBud-olive:
### Changes made:

- Added new Config Aptitude with `getLoopConfig` function
  - The Core returns a JSON string, so the function will be parsing that string to return an object to Loop authors
- Added self tests

### Additional notes:

Merging this branch into my other Loop Config branch since they're needed together in order to work & test

### DIFF
--- a/ldk/javascript/examples/self-test-loop/src/tests/config/index.ts
+++ b/ldk/javascript/examples/self-test-loop/src/tests/config/index.ts
@@ -25,19 +25,17 @@ export const testLoopConfig = (): Promise<boolean> =>
 
           if (JSON.stringify(EXPECTED_VARIABLE_NAMES) === JSON.stringify(actualVariableNames)) {
             console.log('Success: loopConfig keys match expected schema');
-            resolve(true);
+            return resolve(true);
           }
 
-          reject(
+          return reject(
             new Error(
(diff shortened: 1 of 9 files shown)


alexmatthew-olive (approved):

ldk/javascript/examples/self-test-loop/src/tests/config/index.ts
```diff
@@ -0,0 +1,43 @@
...
+            'testString',
+            'testStringArray',
+            'testStringEnum',
+          ];
+
+          const actualVariableNames = Object.keys(loopConfig).sort();
+
+          if (JSON.stringify(EXPECTED_VARIABLE_NAMES) === JSON.stringify(actualVariableNames)) {
+            console.log('Success: loopConfig keys match expected schema');
+            resolve(true);
```
iangallOlive, line 28: I never know if it's **needed**, but I think a `return` after this line would be "good" to include

iangallOlive (approved):

ldk/javascript/examples/self-test-loop/src/tests/config/index.ts
JosefBud-olive: Just for you, buddy

PR #553, opened 2022-04-13
HELPS-1792 Add webpack warning when loop.js does not contain ui.loopO…

racary-olive:
HELPS-1792 Add webpack warning when loop.js does not contain ui.loopOpenHandler

### Changes made:

- Create custom plugin to check if loop.js file contains loopOpenHandler function call

### Additional notes:
image

### DIFF
--- a/ldk/javascript/src/webpack/check-loop-open-handler-plugin.ts
+++ b/ldk/javascript/src/webpack/check-loop-open-handler-plugin.ts
@@ -7,17 +7,11 @@ const plugin = 'CheckLoopOpenHandlerPlugin';
 export class CheckLoopOpenHandlerPlugin {
   apply(compiler: Compiler): void {
     compiler.hooks.thisCompilation.tap(plugin, (compilation) => {
-      compilation.hooks.processAssets.tap({ name: plugin }, (assets) => {
+      compilation.hooks.afterOptimizeAssets.tap({ name: plugin }, (assets) => {
         const loopjs = assets[assetName];
         if (loopjs) {
           const fileContent = loopjs.source();
-          if (
-            !fileContent?.includes('ui.loopOpenHandler(') ||
-            fileContent?.includes('// ui.loopOpenHandler(') ||
-            fileContent?.includes('//ui.loopOpenHandler(') ||
-            fileContent?.includes('* ui.loopOpenHandler(') ||
-            fileContent?.includes('*ui.loopOpenHandler(')
-          ) {
+          if (!fileContent?.includes('ui.loopOpenHandler(')) {
             const openHandlerMissingWarning = new WebpackError(warningMsg);
             openHandlerMissingWarning.file = assetName;
             compilation.warnings.push(openHandlerMissingWarning);
--- a/ldk/javascript/src/webpack/check-loop-open-handler-plugin.ts
+++ b/ldk/javascript/src/webpack/check-loop-open-handler-plugin.ts
@@ -1,0 +1,29 @@
+import { Compiler, WebpackError } from 'webpack';
+
+const assetName = 'loop.js';
+const warningMsg =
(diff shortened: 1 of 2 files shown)


ldk/javascript/src/webpack/check-loop-open-handler-plugin.ts
```diff
@@ -0,0 +1,29 @@
...
+  apply(compiler: Compiler): void {
+    compiler.hooks.thisCompilation.tap(plugin, (compilation) => {
+      compilation.hooks.processAssets.tap({ name: plugin }, (assets) => {
+        const loopjs = assets[assetName];
+        if (loopjs) {
+          const fileContent = loopjs.source();
+          if (
+            !fileContent?.includes('ui.loopOpenHandler(') ||
+            fileContent?.includes('// ui.loopOpenHandler(') ||
+            fileContent?.includes('// ui.loopOpenHandler(') ||
```
iangallOlive, line 17: This is the same as the previous line (unless I'm missing something)

racary-olive: you are right, i meant it to be `fileContent?.includes('//ui.loopOpenHandler(')`. ill update

iangallOlive: Gotcha, thanks for clarifying

ldk/javascript/src/webpack/check-loop-open-handler-plugin.ts
```diff
@@ -0,0 +1,29 @@
...
+  'Please provide a way for users to open your Loop. You can do this using the ui.loopOpenHandler. This is required to submit your Loop.';
+const plugin = 'CheckLoopOpenHandlerPlugin';
+export class CheckLoopOpenHandlerPlugin {
+  apply(compiler: Compiler): void {
+    compiler.hooks.thisCompilation.tap(plugin, (compilation) => {
+      compilation.hooks.processAssets.tap({ name: plugin }, (assets) => {
+        const loopjs = assets[assetName];
+        if (loopjs) {
+          const fileContent = loopjs.source();
+          if (
```
iangallOlive, line 14: I'm sure this is better than what we have now, I just want to point out some scenarios that this might handle correctly:

- The loop.js file is not minimized
- Some optimization that renames `ui` to something else
- Console log messages that have this string in them
- Commented out code that has an uncommented version that *does* make this call

racary-olive: @iangallOlive those are all good/valid scenarios. I think the last statement is the most likely as where we intercept the loop js file at is not in its fully optimized form.

I tried to find a better solution but was unable to unless we wanted to read over each file in a project (didnt think that was great either)

I can try to update the logic to handle that last scenario more smoothly

racary-olive: I can change to use this `compilation.hooks.afterOptimizeAssets` this hook will give us access to the optimized loop js file which does not include any comments and we could just loop for `loopOpenHandler(`

iangallOlive: Yeah, and I wasn't trying to criticize your work, just that it's a really hard problem to fully solve.

I think checking after optimization is a great idea, and might solve everything. My only question is: what happens if they *don't* optimize? I know they're supposed to import and use the webpack config that the LDK provides. But like what if they set their webpack config to not minify (or disable whatever other optimization)? I do believe I've seen something like this before - the dev was doing something different/special when in development mode around optimization. Still, we probably don't care about `development` mode giving a false negative or something.

racary-olive: yeah thats another good point. We could have both hooks in place as an ultimate fail safe but assume (probably dangerous) that users will optimize

iangallOlive (approved):

jeffsun-olive (approved):

PR #558, opened 2022-04-14
Helps 4131 listen tab change

racary-olive:
### Changes made:

- Create TabChangeDetails interface
- Add listenTabChange function to browser interface
- add new Browser test to self test loop

### Additional notes:
image

### DIFF
--- a/ldk/javascript/examples/self-test-loop/src/tests/browser/index.ts
+++ b/ldk/javascript/examples/self-test-loop/src/tests/browser/index.ts
@@ -159,12 +159,13 @@ export const testTabChangeEvent = (): Promise<boolean> =>
           listener.cancel();
           resolve(true);
         }
-
-        setTimeout(() => {
-          // After 3 seconds, if tabs havent been switched throw error
-          reject(new Error('The text selection in your browser does not match the test text'));
-        }, 3000);
       });
+
+      setTimeout(() => {
+        listener.cancel();
+        // After 3 seconds, if tabs havent been switched throw error
+        reject(new Error('Tab change event took longer than allotted time'));
+      }, 3000);
     } catch (error) {
       console.error(error);
       reject(error);
--- a/ldk/javascript/examples/self-test-loop/src/tests/browser/index.ts
+++ b/ldk/javascript/examples/self-test-loop/src/tests/browser/index.ts
@@ -7,6 +7,7 @@ import {
   TabChangeDetails,
   UIElementDetails,
 } from '@oliveai/ldk/dist/browser';
+import { Cancellable } from '../../../../../dist/cancellable';
 
 export const testOpenTabAndListenNavigation = (): Promise<boolean> =>
   new Promise(async (resolve, reject) => {
@@ -145,6 +146,9 @@ export const testOpenWindowAndListenNavigation = (): Promise<boolean> =>
 
 export const testTabChangeEvent = (): Promise<boolean> =>
   new Promise(async (resolve, reject) => {
+    let listener: void | Cancellable;
+    let timeout;
+
     try {
       const URL = 'https://www.oliveai.dev/';
       const URL2 = 'https://www.oliveai.com/';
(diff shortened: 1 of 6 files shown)


ldk/javascript/examples/self-test-loop/src/tests/browser/index.ts
```diff
@@ -143,6 +143,34 @@ export const testOpenWindowAndListenNavigation = (): Promise<boolean> =>
     }
   });
 
+export const testTabChangeEvent = (): Promise<boolean> =>
```
iangallOlive, line 146: I have no idea if this is true, or if there's any issues around it, but I feel like these self tests don't usually do a good job cleaning up listeners/timeouts. You don't have to change anything, but this seems to manage things "better":

```javascript
export const testTabChangeEvent = (): Promise<boolean> =>
  new Promise(async (resolve, reject) => {
    let listener;
    let timeout;

    try {
      const URL = 'https://www.oliveai.dev/';
      const URL2 = 'https://www.oliveai.com/';
      const TITLE = 'Olive Helps Developer Hub';

      await browser.openTab(URL);
      await browser.openTab(URL2);

      await Promise.race([
        new Promise((res2, rej2) => {
          listener = browser.listenTabChange((tabChangeDetails: TabChangeDetails): void => {
            const { tabId, title, url, windowId } = tabChangeDetails;
            if (tabId && windowId && title === TITLE && url === URL) {
              res2(true);
            }
          }).catch((err) => {
            rej2(err);
          });
        }),
        new Promise((_, rej2) => {
          // After 3 seconds, if tabs haven't been switched throw error
          timeout = setTimeout(() => rej2(new Error('The text selection in your browser does not match the test text')), 3000);
        }),
      ]);
    } catch (error) {
      console.error(error);
      reject(error);
    } finally {
      if (listener) {
        listener.cancel();
      }
      if (timeout) {
        clearTimeout(timeout);
      }
    }
  });
```

ldk/javascript/examples/self-test-loop/src/tests/browser/index.ts
```diff
@@ -143,6 +143,34 @@ export const testOpenWindowAndListenNavigation = (): Promise<boolean> =>
...
+      const listener = await browser.listenTabChange((tabChangeDetails: TabChangeDetails): void => {
+        const { tabId, title, url, windowId } = tabChangeDetails;
+        if (tabId && windowId && title === TITLE && url === URL) {
+          listener.cancel();
+          resolve(true);
+        }
+
+        setTimeout(() => {
+          // After 3 seconds, if tabs havent been switched throw error
+          reject(new Error('The text selection in your browser does not match the test text'));
```
iangallOlive, line 165: I think this error message needs to be updated

ldk/javascript/examples/self-test-loop/src/tests/browser/index.ts
iangallOlive: The scenarios where I think this "improves" the existing test structure:

- Say the user starts this test, but they manually open some new tabs (which triggers the `listenTabChange` callback, which calls a *new* `setTimeout`), and they switch around the tabs (never to the first one, which would resolve the test). A bunch of `setTimeout`s are called and never cleared (which is basically fine, because they will after 3 seconds, redundantly calling `reject`). But still, it feels like we can avoid/handle this better.
- Say the user starts this test, and they never switch tabs (which fails the test). The `listener` never gets cancelled. This seems more important.

iangallOlive (approved):

ldk/javascript/examples/self-test-loop/src/tests/browser/index.ts
racary-olive: yeah i agree listener cleanup was not great. I think using Promise.race in this case isnt needed since we dont want to race against the listener being registered, we just want to make sure the tab event is happeneing within the 3 seconds

I also found tha tusing finally causes issues becuase the finally can be called before the tab event has come through the listener callback.

I did make some updates to make sure were cancelling the listener even if there is an error

ldk/javascript/examples/self-test-loop/src/tests/browser/index.ts
racary-olive: update message

ldk/javascript/examples/self-test-loop/src/tests/browser/index.ts
racary-olive: I also tried making the test fully automate. Unfortunately the tab event that gets created by openTab is different than when an actual tab `switch` happens (not sure why that is)

iangallOlive: That sounds good about being "fully automated" - thanks for checking!

And maybe I'm missing the point, or wrote it wrong (I didn't run/test that code anywhere), but I thought the test was essentially "open the tabs, wait for the user to switch to the first tab (within 3 seconds)".

And I **thought** `await`ing the `Promise.race` basically did that. So it creates both tabs, then it starts the listener and the 3-second timeout, and awaiting the race basically accomplished "wait until whatever happens first - either the correct tab change occurring OR the 3-second timeout". But again, I could either be misunderstanding, or wrote it wrong.

Either way, thanks for addressing the listener cleanup!

iangallOlive: For what it's worth, a simple example seems to work how I expected it to: https://jsfiddle.net/oxsdL1z7/

racary-olive: you're right!  had to update the reject/resolve thats being called within the race condition for it to function properly within the self test loop.

iangallOlive: Gotcha, I think I realized one aspect was not fully correct in my original example and edited the comment. But either way, thanks for entertaining my suggestion and glad you got it working. I was just moreso interested to see how the test could look if it was reorganized to "better" handle these things; you're always welcome to argue against my ideas!

racary-olive: yeah i think that makes sense. It took me a minute to remember how promises & reject/resolve works especially with nested promises. I'm definitely all in for clean code!

iangallOlive: And I appreciate it!

As I look at the function more, I can definitely see confusion due to nesting promises and mixing promises with async/await at different levels. So those things could definitely be improved, but I feel like the latest functionality is more solid at this point, and syntax/clarity is what's left.

jeffsun-olive (approved):

PR #566, opened 2022-04-21
HELPS-4347 return after reject

racary-olive:
### Changes made:

- Add return statement after rejecting a promise to stop function execution
- Add test cases for network websocket


harrisonkunkel (approved):

ldk/javascript/src/network/mapper.ts
```diff
@@ -121,6 +121,7 @@ export const mapToSocket = (socket: Network.Socket): Socket => ({
           if (error) {
             console.error(`Received error on ping ${error.message}`);
             reject(error);
+            return;
```
iangallOlive, line 124: Good find!

iangallOlive (approved):

PR #567, opened 2022-04-21
added getEnvironment method to system aptitude

harrisonkunkel:
### Changes made:

- added method getEnvironment() to system aptitude
- added  new test to self test loop for getEnvironment 
- updated test for operatingSystem()
- added unit tests 

### Additional notes:
Screen Shot 2022-04-22 at 12 22 59 PM

Screen Shot 2022-04-22 at 12 23 38 PM

### DIFF
--- a/ldk/javascript/examples/self-test-loop/src/tests/system/index.ts
+++ b/ldk/javascript/examples/self-test-loop/src/tests/system/index.ts
@@ -14,3 +14,24 @@ export const testOperatingSystem = (): Promise<boolean> =>
         reject(error);
       });
   });
+
+export const testGetEnvironment = (): Promise<boolean> =>
+  new Promise((resolve, reject) => {
+    system
+      .getEnvironment()
+      .then((getEnvironment) => {
+        console.debug(`System Aptitude: getEnvironment: osVersion: ${getEnvironment.osVersion}`);
+        console.debug(
+          `System Aptitude: getEnvironment: oliveHelpsVersion: ${getEnvironment.oliveHelpsVersion}`,
+        );
+        console.debug(
+          `System Aptitude: getEnvironment: loopVersion: ${getEnvironment.loopVersion}`,
+        );
+        setTimeout(() => {
+          resolve(true);
+        }, 1000);
+      })
+      .catch((error) => {
+        reject(error);
+      });
+  });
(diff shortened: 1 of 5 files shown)


ldk/javascript/examples/self-test-loop/src/tests/system/index.ts
```diff
@@ -14,3 +14,24 @@ export const testOperatingSystem = (): Promise<boolean> =>
...
+      .getEnvironment()
+      .then((getEnvironment) => {
+        console.debug(`System Aptitude: getEnvironment: osVersion: ${getEnvironment.osVersion}`);
+        console.debug(
+          `System Aptitude: getEnvironment: oliveHelpsVersion: ${getEnvironment.oliveHelpsVersion}`,
+        );
+        console.debug(
+          `System Aptitude: getEnvironment: loopVersion: ${getEnvironment.loopVersion}`,
+        );
+        setTimeout(() => {
```
racary-olive, line 30: instead of just resolving true after a second could we add a check for each property and if each check passes then we resolve true.

I think the setTimeout should reject if the test doesnt pass within a certain amount of time

ldk/javascript/src/system/system.test.ts
```diff
@@ -32,4 +33,34 @@ describe('System', () => {
...
+      };
+      mocked(oliveHelps.system.getEnvironment).mockImplementation((callback) =>
+        callback(undefined, environment),
+      );
+
+      const actual = system.getEnvironment();
+
+      expect(oliveHelps.system.getEnvironment).toHaveBeenCalledWith(expect.any(Function));
+
+      return expect(actual).resolves.toBe(environment);
```
racary-olive, line 52: I dont think we need to return anything here

ldk/javascript/src/system/system.test.ts
```diff
@@ -32,4 +33,34 @@ describe('System', () => {
...
+
+    it('returns a rejected promise', () => {
+      const exception = 'Exception';
+      mocked(oliveHelps.system.getEnvironment).mockImplementation(() => {
+        throw exception;
+      });
+
+      const actual = system.getEnvironment();
+
+      return expect(actual).rejects.toBe(exception);
```
racary-olive, line 63: same comment as above

ldk/javascript/examples/self-test-loop/src/tests/system/index.ts
jeffsun-olive: Instead of display the message in console, can we create a whisper and show `osVersion`, `oliveHelpsVersion` and `loopVersion` there and provide buttons for `pass` or `fail`, something like this:
```
     <oh-box
            direction={whisper.Direction.Horizontal}
            justifyContent={whisper.JustifyContent.SpaceEvenly}
          >
            <oh-button label="Pass" onClick={() => resolve(true)} />
            <oh-button label="Fail" onClick={() => reject()} />
          </oh-box>

jeffsun-olive (approved):

racary-olive (approved):

PR #572, opened 2022-04-27
Helps 4211 UI listener type

racary-olive:
### Changes made:

- Add listenerType to UIElementArguments type definition

### DIFF
--- a/ldk/javascript/src/types/oliveHelps/browser.d.ts
+++ b/ldk/javascript/src/types/oliveHelps/browser.d.ts
@@ -48,9 +48,10 @@ declare namespace Browser {
     windowId: number;
   }
 
+  export type ListenType = 'click' | 'input';
   interface UIElementArguments {
     address: string;
-    listenerType?: string;
+    listenerType?: ListenType;
     selector: string;
   }
   interface UIElementDetails {
--- a/ldk/javascript/src/types/oliveHelps/browser.d.ts
+++ b/ldk/javascript/src/types/oliveHelps/browser.d.ts
@@ -49,8 +49,9 @@ declare namespace Browser {
   }
 
   interface UIElementArguments {
-    selector: string;
     address: string;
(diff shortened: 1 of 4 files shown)


ldk/javascript/src/types/oliveHelps/browser.d.ts
```diff
@@ -49,8 +49,9 @@ declare namespace Browser {
   }
 
   interface UIElementArguments {
-    selector: string;
     address: string;
+    listenerType?: string;
```
MaggieMaSS, line 53: if `listenerType` is limited to certain options, instead of using string directly, we usually create a type `ListenerTypes` and the prop would look like this: 
`listenerType?: ListenType; `

racary-olive: good call, made the updates

MaggieMaSS (approved):

harrisonkunkel:
Could you update the self test loop?

harrisonkunkel (approved):

ldk/javascript/src/types/oliveHelps/browser.d.ts
```diff
@@ -48,12 +48,15 @@ declare namespace Browser {
     windowId: number;
   }
 
+  export type ListenType = 'click' | 'input';
```
jeffsun-olive, line 51: I get a dummy question about the ListenType. Do we only listen to these two types? Should we discuss what need to be added to the list and what should not? I think there are many different types of events on Browser.

racary-olive: Thats a great point. There are definitely other event types that can be listened to. These were the only two that have seemed to be requested so far. Heres a list of [event](https://developer.mozilla.org/en-US/docs/Web/Events) that we could potentially account for/add to this list

PR #586, opened 2022-05-11
Helps 4417 restructure

racary-olive:
### Changes made:

-  Restructure LDK

### Additional notes:
This will require updating all of the documentation links from `/ldk/javascript/...` to `/...` for `interface`, `enumerations` & `type aliases`. 

We will also need to update the examples link found here https://app.gitbook.com/o/-MefJkFm4JRcrZoWaBy4/s/-MfdLaju6LfXozdWGii8/ldk/examples


iangallOlive:
This should point to `develop`, right?

racary-olive:
@iangallOlive yes and thank you!

iangallOlive:
Overall, this looks exactly as I hoped, so great work!

There's only a few tiny things I noticed:
- In `.github/workflows/docs.yml`, I **think** the references to the `javascript` dir can be removed from the `# Consolidate docs` section.
- In `.gitattributes`, you can remove the lines for `*.go` and `*.cs`.
- In `LICENSE`, you can update the copyright year to be `2022`.
- Rather than have the `.github/workflows/prettier.yml` file, you could just add `npm run prettier:all:check` to the `.github/workflows/javascript.yml` file (like right after the `- run: npm run lint` line).
- In `.eslintignore`, we should include an entry for `docs`
- In `.github/workflows/javascript.yml`, I think we want the `- run: npm run build --if-present` to be at the end of the list of the `steps`, and I don't think `- run: npm run doc` is needed (unless you can think of a reason)

iangallOlive (approved):

iangallOlive:
Also, I know we decided to remove the whole `docs` directory. But one important file inside of there is the public key that's needed to verify JWTs created by the User aptitude (https://github.com/open-olive/loop-development-kit/blob/main/docs/static/public-keys/olive-helps/user-aptitude/user-aptitude-jwt-rsa.pub). We link to it on the Dev site, and it's **possible** that a third-party service references the file by that exact path. So if we were to remove it, **something** _could_ break. I would recommend we restore that file in this PR, for now. It's unfortunate we stored it in this repo, especially under this path, but I don't have a better way to handle it at this point, so we should just keep it there for now.

jeffsun-olive (approved):

PR #597, opened 2022-05-18
listen2 function

harrisonkunkel:
### Changes made:
- changed listen function to use listen2
- removed listen and includeOliveHelpsEvents from clipboard Aptitude

### Additional notes:
- no STL changes needed

### DIFF
--- a/ldk/javascript/src/clipboard/index.ts
+++ b/ldk/javascript/src/clipboard/index.ts
@@ -1,5 +1,5 @@
 import { Cancellable } from '../cancellable';
-import { promisify, promisifyListenable, promisifyWithParam } from '../promisify';
+import { promisify, promisifyWithParam, promisifyListenableWithParam} from '../promisify';
 
 /**
  *  The Clipboard aptitude provides access to the OS's clipboard.
@@ -33,8 +33,7 @@ export function listen(
   includeOliveHelpsEvents: boolean,
   callback: (clipboardText: string) => void,
 ): Promise<Cancellable> {
-  oliveHelps.clipboard.includeOliveHelpsEvents(includeOliveHelpsEvents);
-  return promisifyListenable(callback, oliveHelps.clipboard.listen);
+  return promisifyListenableWithParam(includeOliveHelpsEvents, callback, oliveHelps.clipboard.listen2);
 }
 
 export function read(): Promise<string> {
--- a/ldk/javascript/src/types/oliveHelps/clipboard.d.ts
+++ b/ldk/javascript/src/types/oliveHelps/clipboard.d.ts
@@ -4,8 +4,10 @@ declare namespace Clipboard {
 
     write: Common.ReadableWithParam<string, void>;
 
-    listen: Common.Listenable<string>;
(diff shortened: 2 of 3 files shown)


ldk/javascript/src/clipboard/index.ts
```diff
@@ -20,21 +20,60 @@ export interface Clipboard {
...
+
+      const setFocusedWindow = () =>
+        new Promise((res, rej) => {
+          oliveHelps.window.activeWindow((error, event) => {
+            if (error) {
+              return rej(error);
+            }
+            focusedWindow = event.title;
+
+            return res(focusedWindow);
```
racary-olive, line 52: since we no longer are setting focusedWindow in multiple places i think it might make sense to remove line 50 & just do `return resolve(event.title)`

Then in line 61 do this

`const focusedWindow = await setFocusedWIndow()`

This way we dont have to deal with a global variable

ldk/javascript/src/clipboard/index.ts
```diff
@@ -20,21 +20,60 @@ export interface Clipboard {
...
+          });
+        });
+
+      oliveHelps.clipboard.listenAll(
+        async (error, event) => {
+          if (error) {
+            return error;
```
racary-olive, line 59: I think we want to reject the error

ldk/javascript/src/types/oliveHelps/clipboard.d.ts
```diff
@@ -4,8 +4,6 @@ declare namespace Clipboard {
 
     write: Common.ReadableWithParam<string, void>;
 
-    listen: Common.Listenable<string>;
-
-    includeOliveHelpsEvents(enabled: boolean): void;
+    listenAll: Common.Listenable<string>;
```
racary-olive, line 7: should this be listen?

jeffsun-olive: I think the new function we had in `core` is called `listenAll`, so it is correct.

ldk/javascript/src/clipboard/index.ts
harrisonkunkel: Do it not do that already? it is in a try catch statement that if there is an error then it reject(error). Line 74

ldk/javascript/src/types/oliveHelps/clipboard.d.ts
racary-olive: my mistake, didnt look at the folder the file was in!

ldk/javascript/src/clipboard/index.ts
racary-olive: I wasnt sure if this error would be caught in the catch block or not since its a part of the inner callback but i think your right

ldk/javascript/src/clipboard/index.ts
harrisonkunkel: done

racary-olive (approved):

MaggieMaSS (approved):

PR #600, opened 2022-06-03
HELPS-4739-autocomplete-filter-props-mutually-exclusive

harrisonkunkel:
### Changes made:

- added types AutocompleteFilter and AutocompleteMatchSorter
- type Autocomplete can only have one of the types above
- types have either matchSorter or filterOptions props not both.


src/whisper/types.ts
```diff
@@ -558,6 +543,28 @@ export type Autocomplete = SelectComponent<WhisperComponentType.Autocomplete> &
...
+export type AutocompleteMatchSorter = {
+  /**
+   * Options to use with our implementation of match-sorter
+   * https://github.com/kentcdodds/match-sorter
+   *
+   * (Note: filterOptions and matchSorter can't be used together; if you include both,
+   * matchSorter will take precedence)
+   */
+  matchSorter?: AutocompleteMatchSorterOptions;
+  filterOptions?: never;
```
racary-olive, line 567: should filterOptions be type AutocompleteMatchSorterOptions instead of matchSorter?

racary-olive (approved):

jeffsun-olive (approved):

PR #601, opened 2022-06-07
add writemessage to websocket selftest

jeffsun-olive:
### Changes made:

- update Network Aptitude -Websocket self test


ldk/javascript/examples/self-test-loop/src/tests/network/index.ts
```diff
@@ -155,25 +143,25 @@ export const testWebsocketConnection = (): Promise<boolean> =>
...
+            console.error('setCloseHandler error', error);
+            reject(error);
+          }
+
+          console.info(`Received on close code: ${code}. ${text}`);
+
+          // This is from us closing the socket connection in the message handler below
+          if (testPassed) {
+            resolve(true);
+          }
```
harrisonkunkel, line 160: why dont you just do something like 

```
if (message === 'You are using a test api key') {
  resolve(true);
}
```
and get rid of the testPassed?

ldk/javascript/examples/self-test-loop/src/tests/network/index.ts
```diff
@@ -121,31 +121,19 @@ export const testWebsocketConnection = (): Promise<boolean> =>
...
-
-          console.info(`Received on close code: ${code}. ${text}`);
-
-          // This is from us closing the socket connection in the message handler below
-          if (testPassed) {
-            resolve(true);
-          }
-        });
-
+        socket.writeMessage('You are using a test api key');
```
racary-olive, line 124: I think we should do `await socket.writeMessage('You are using a test api key');` since `writeMessage` returns a `Promise`

ldk/javascript/examples/self-test-loop/src/tests/network/index.ts
```diff
@@ -155,25 +143,25 @@ export const testWebsocketConnection = (): Promise<boolean> =>
 
           clipboardListener.cancel();
           console.info('Clipboard listener cancelled');
-
           cancellable.cancel();
         });
+
+        await socket.setCloseHandler((error, code, text) => {
```
racary-olive, line 149: Im wondering if we should do this before we do `const cancellable = await socket.setMessageHandler` I'm not sure if setCloseHandler is actually being called since were awaiting the setMessageHandler and the first thing were doing is writing a message

racary-olive: Im also wondering if the very first thing we should do is call `setMessageHandler`

ldk/javascript/examples/self-test-loop/src/tests/network/index.ts
harrisonkunkel: does setCloseHandler not have access to message? I thought it would, but if not just ignore me.

harrisonkunkel (approved):

ldk/javascript/examples/self-test-loop/src/tests/network/index.ts
jeffsun-olive: If I put setCloseHandler before all the other handlers, it will stuck in setCloseHandler, since we are not calling socket close yet. 
I have tested switch setMessageHandler and WriteMessage doesn't affect the test result. I would move setMessageHandler be the first thing to call.

MaggieMaSS (approved):

